Issue #1, opened 2017-02-24
Tensor Structure Definition


tqchen:
migrated from https://github.com/dmlc/mxnet/issues/4735

tqchen:
I take a stab in making the basic tensor structure here https://github.com/dmlc/mxnet/issues/4735 based on existing discussions

tqchen:
I would like to open a vote discussion for current data structure in dlpack.h, please voice your opinion or simply vote by +1  if you agree.  If there is no objection in the proposal in one week, we will consider it finalized. @soumith @edgarriba @bhack @piiswrong @Yangqing

Yangqing:
Thanks @tqchen ! It looks great in general, a few minor comments:

- What is kCPUPinned? I understand the implication (i.e. cudaMallocHost) but it seems that the implementation won't distinguish kCPU vs kCPUPinned. A bit more documentation would be great.

- For DLTypeCodeKind, it might be good to explicitly specify the length, like int8_t, int16_t, int32_t and int64_t. Using int and long has been a headache because compilers have different ways of dealing with them, even templating (see https://github.com/caffe2/caffe2/blob/master/cmake/MiscCheck.cmake#L5-L22)

- For DLContext - did you intend to use enum instead of int?

- For DLTensor - I like the opaque pointer! This has been something I was very vocal about but not that well accepted by some frameworks.

tqchen:
@Yangqing One thing that hangs on the DLDataType, is to make it future compatibile with weird types, like int1_t.  

But you are right that having such thing makes it a bit hard for templatizing, maybe we can consider use bit packing to pack the code and bits into a single int.

bhack:
Taking a look to [TF data types](https://www.tensorflow.org/programmers_guide/dims_types) seems that bool could be useful. What about quantized?

tqchen:
The current data type cane expression quantized data types like ```int4_t```,  by specifying bits=4 and type_code=int.

I think what @Yangqing mentioned is being able to use type constant as template argument, using a  struct kinda of goes against that. Maybe using  a bit packed integer(with same layout as struct) helps to resolve that problem

tqchen:
@Yangqing Consider open a PR with respect your suggested changes?

Yangqing:
@tqchen Might be a bit tricky due to my day duties but I'll try at my earliest availability :)

Yangqing:
[...]

PR #8, opened 2017-02-28
minor tuning

Yangqing:
This basically names things a little bit consistent with the comments, and used enum types whenever applicable for safety. Will put some more comments.

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -36,16 +36,7 @@ typedef enum {
   // kCPUPinned = kCPU | kGPU
   kCPUPinned = 3,
   kOpenCL = 4,
-} DLDeviceKind;
-
-/*!
- * \brief The type code options DLDataType.
- */
-typedef enum {
-  kInt = 0U,
-  kUInt = 1U,
-  kFloat = 2U
-} DLTypeCodeKind;
+} DLDeviceType;
 
 /*!
  * \brief A Device context for Tensor and operator.
@@ -54,8 +45,17 @@ typedef struct {
   /*! \brief The device index */
   int device_id;
   /*! \brief The device type used in the device. */
-  int device_type;
+  DLDeviceType device_type;
 } DLContext;
+
+/*!
+ * \brief The type code options DLDataType.
+ */
+typedef enum {
+  kInt = 0,
+  kUInt = 1,
+  kFloat = 2,
+} DLDataTypeCode;
 
 /*!
  * \brief The data type the tensor can hold.
@@ -67,7 +67,7 @@ typedef struct {
  */
 typedef struct {
   /*! \brief Type code of base types */
-  uint8_t code;
+  DLDataTypeCode code;
   /*!
    * \brief Number of bits, common choices are 8, 16, 32.
    */
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -52,9 +52,9 @@ typedef struct {
  * \brief The type code options DLDataType.
  */
 typedef enum {
-  kInt = 0,
-  kUInt = 1,
-  kFloat = 2,
+  kInt = 0U,
+  kUInt = 1U,
+  kFloat = 2U,
 } DLDataTypeCode;
 
 /*!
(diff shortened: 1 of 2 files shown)


include/dlpack/dlpack.h
```diff
@@ -54,10 +45,19 @@ typedef struct {
...
 } DLContext;
 
 /*!
+ * \brief The type code options DLDataType.
+ */
+typedef enum {
+  kInt = 0,
+  kUInt = 1,
+  kFloat = 2,
+} DLDataTypeCode;
```
Yangqing: (I renamed from DLTypeCodeKind to DLDataTypeCode - it is the code used in DLDataType, which is clearer.)

include/dlpack/dlpack.h
```diff
@@ -67,7 +67,7 @@ typedef struct {
  */
 typedef struct {
   /*! \brief Type code of base types */
-  uint8_t code;
+  DLDataTypeCode code;
```
Yangqing: So, at a high level, I have some doubts about the complexity of the DLDataType. For example, what if someone marks DLDataTypeCode as kFloat but then sets the bits to, say, 17?

I actually like the current design of TensorFlow and Caffe2 - in the sense that we explicitly list out the data types instead of using bits. The reason is that there are not much possible choices of the bits value, so it is easier to define FLOAT64, FLOAT32, FLOAT16 instead of doing error-prone (kFloat, 23). 

By the way, I still like the existence of DLDataType because of lanes - I think we should keep it, but kill the bits field. Thoughts?

Yangqing:
#1

tqchen:
I still hope we could be future compatible for things like 1 bit, or 17 bits like you mentioned for serialization. How about

```c++
enum int {
   kInt = 1;
   kUInt = 2;
   kFloat= 3;
} DLDataTypeKind;

enum : uint16_t {
  kFloat32 = (kFloat << 8U) | 32U;
  kInt8 = (kInt << 8U) | 8U;
} DLDataTypeCode;
```
In this way, user always choose from the typical type codes, enables templatizing by DLDataTypeCode, but still the enum encoding can encode the bit fields.

Yangqing:
I don't have a strong opinion either for or against - both will serve the purpose. BTW I am not that worried about templates, because we can always do something like

```
template <int code, int width>
class RealType;
```

and specialize RealType<kInt, 32>::type for templates.

include/dlpack/dlpack.h
tqchen: The reason to keep it as uint8_t is because enum is going to become int, which makes it 8 bytes, while original field packs into 4 bytes

tqchen: Unfortunately enum with type is feature of c++11, but not c...

include/dlpack/dlpack.h
```diff
@@ -54,10 +45,19 @@ typedef struct {
   /*! \brief The device index */
   int device_id;
   /*! \brief The device type used in the device. */
-  int device_type;
+  DLDeviceType device_type;
```
tqchen: agree with this chance

include/dlpack/dlpack.h
Yangqing: Good point.

tqchen:
I see. I think there are two needs:

- Possible need of being future compatible with low bits
- Only list out possible choice options (like kFloat32, kInt8)

I think the bit field packing convention could be reasonable to serve both purposes, and there is minimum overhead if only non-conventional bits are used (user choose from predefined TypeCodes)

include/dlpack/dlpack.h
tqchen: keep the 'u' suffix so it is unsigned, as the code is unsigned, to prevent compiler warning

Yangqing:
Yeah I agree - makes sense to keep the current kInt/kUint/kFloat + a bit length setting.

tqchen:
I think it is good to merge, if you don't have future changes

include/dlpack/dlpack.h
Yangqing: Changed back. I am curious though - the standard does not specify if an enum is signed or unsigned. Which compiler is producing the error? And is it -Wsign-compare?

Yangqing:
Cool, that was quick, thanks @tqchen

PR #14, opened 2017-06-01
[BUILD/TEST] Formalize tests, project structure

tqchen:
- Add travis tests to make sure the style and compilation pass
- Formalize project structure to include stable components in include and others in contrib

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -3,14 +3,17 @@
  * \file dlpack.h
  * \brief The common header of DLPack.
  */
-#ifndef DLPACK_H_
-#define DLPACK_H_
+#ifndef DLPACK_DLPACK_H_
+#define DLPACK_DLPACK_H_
 
 #ifdef __cplusplus
 #define DLPACK_EXTERN_C extern "C"
 #else
 #define DLPACK_EXTERN_C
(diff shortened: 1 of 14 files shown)


include/dlpack/dlpack.h
```diff
@@ -3,15 +3,18 @@
...
+#define DLPACK_DLPACK_H_
 
 #ifdef __cplusplus
 #define DLPACK_EXTERN_C extern "C"
 #else
 #define DLPACK_EXTERN_C
 #endif
 
+/*! \brief The current version of dlpack */
+#define DLPACK_VERSION 010
```
edgarriba: version number should go accord to
https://github.com/tqchen/dlpack/blob/master/CMakeLists.txt#L11

tqchen: Updated in the cmake list. is that good?

edgarriba: yeah, but not sure what's the proper way to have a centralized version registry

tqchen: AFIK we have to do it in both side

tqchen: @edgarriba Now that this is resolved, can we go ahead and merge this in?

PR #16, opened 2017-08-28
[Device Type] Metal Device type

antinucleon:


eric-wieser:
I'm curious - how can the verilog VPI be used to actually get a chunk of memory suitable for use as a dlpack buffer?

PR #17, opened 2017-10-23
New Struct for managing memory of DLTensor

prigoyal:
We recently incorporated using dlpack tensor conversions in PyTorch and also needed to do memory management. In order to do that, we introduce a new struct called `DLManagedTensor` which can hold a nullable pointer to the destructor function that users can define as they wish. More fields of the struct have been explained in the PR and the example usage of this struct can be found here:

https://github.com/zdevito/ATen/blob/master/src/ATen/DLConvertor.cpp

and PyTorch tests for dlpack conversions https://github.com/pytorch/pytorch/blob/master/test/test_torch.py#L4250

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -114,6 +114,27 @@ typedef struct {
   /*! \brief The offset in bytes to the beginning pointer to data */
   uint64_t byte_offset;
 } DLTensor;
+
+/*!
+ * \brief C Tensor object, manage memory of DLTensor. This data structure is
+ *  intended to faciliate the borrowing of DLTensor by another framework. It is
+ *  not meant to transfer the tensor. When the borrowing framework doesn't need
+ *  the tensor, it should call the deleter to notify the host that the resource
+ *  is no longer needed.
+ */
+typedef struct DLManagedTensor {
+  /*! \brief DLTensor which is being memory managed */
+  DLTensor dl_tensor;
+  /*! \brief the context of the original host framework of DLManagedTensor in
+   *   which DLManagedTensor is used in the framework. It can also be NULL.
+   */
+  void * manager_ctx;
+  /*! \brief Destructor signature void (*)(void*) - this should be called
+   *   to destruct manager_ctx which holds the DLManagedTensor. It can be NULL
+   *   if there is no way for the caller to provide a reasonable destructor.
+   */
+  void (*deleter)(struct DLManagedTensor * self);
+} DLManagedTensor;
 #ifdef __cplusplus
 }  // DLPACK_EXTERN_C
 #endif


PENGUINLIONG:
Why this couldn't be done at a higher level?

In C++ we have `unique_ptr`, whose destructors can be customised. In other OOP languages we always have destructors. And defining a callback without specifying call convention.

And I don't think a 'nullptr' destructor is a good idea. If a piece of memory cannot be released properly, it shouldn't be allocated.

tqchen:
since it is a major change proposal I opened an issue here https://github.com/dmlc/dlpack/issues/18

tqchen:
There are some updated comments in the RFC issue, please incorporate those. The travis CI is reporting error due to wrong doxygen comments. You can reproduce by ```make doc|grep warn```

prigoyal:
thanks @tqchen I'll update the PR soon

include/dlpack/dlpack.h
```diff
@@ -114,6 +114,26 @@ typedef struct {
...
   uint64_t byte_offset;
 } DLTensor;
+
+/*!
+ * \brief C Tensor object, manage memory of DLTensor. This data structure is
+ *  intended to faciliate the borrowing of DLTensor by another framework. It is
+ *  not meant to transfer the tensor. When the borrowing framework doesn't need
+ *  the tensor, it should call the deleter to handle reference counting.
+ */
+struct DLManagedTensor {
```
tqchen: We still want to keep typedef style. Otherwise, the reference to DLManagedTensor always has to be ```struct DLManagedTensor``` in C

include/dlpack/dlpack.h
```diff
@@ -114,6 +114,26 @@ typedef struct {
   /*! \brief The offset in bytes to the beginning pointer to data */
   uint64_t byte_offset;
 } DLTensor;
+
+/*!
+ * \brief C Tensor object, manage memory of DLTensor. This data structure is
+ *  intended to faciliate the borrowing of DLTensor by another framework. It is
+ *  not meant to transfer the tensor. When the borrowing framework doesn't need
+ *  the tensor, it should call the deleter to handle reference counting.
```
tqchen: ```to handle reference counting``` -> ```to notify the host that the resource is no longer needed.```

include/dlpack/dlpack.h
```diff
@@ -114,6 +114,26 @@ typedef struct {
...
+ */
+struct DLManagedTensor {
+  /*! \brief DLTensor which is being memory managed */
+  DLTensor dl_tensor;
+  /*! \brief context in which DLManagedTensor is used in a framework. It can
```
tqchen: the context of the original host framework(to emphasize it is the original host) of DLManagedTensor

include/dlpack/dlpack.h
tqchen: i.e. there always has to be a struct prefix

prigoyal:
one of the builds for ios seems to be stuck for long time and hasn't started yet. is there some issue with the CI on ios?

include/dlpack/dlpack.h
tqchen: consider add ```typedef struct DLManagedTensor DLManagedTensor;```

tqchen:
It is simply due to the travis's iOS pending queue is longer than linux ones. It should start eventually

prigoyal:
@tqchen the build is fine, can this be merged?

tqchen:
Thanks! this is merged

PR #26, opened 2018-06-15
add storage type support

liangfu:


tqchen:
As the standardization effort involves multiple packages and careful consensus. Please send an RFC first that elaborate what do we want from here, and we need sufficient discussion before creating PR to merge things in. Thanks!

PR #38, opened 2019-03-28
Add an example for borrowing numpy arrays without copying the content

junrushao1994:
This PR gives an example that we could do "numpy.ndarray -> DLManagedTensor" with zero-copy.


apps/from_numpy/main.c
```diff
@@ -0,0 +1,45 @@
+#include <stdio.h>
+#include <dlpack/dlpack.h>
+
+DLManagedTensor given;
+
+void display(DLManagedTensor a) {
```
tqchen: change main.c -> numpy_dlpack.c

junrushao1994:
@tqchen Updated accordingly. Could you offer another round of review? Thanks!

Issue #39, opened 2019-05-21
[ambiguity] Does DLManagedTensor::deleter delete its function argument?


junrushao1994:
It seems to me an undefined behavior in our protocol whether the deleter below deletes `self` as well. The inconsistency causes a recent crash when using MXNet as a backend for DGL, reported by @zheng-da.

https://github.com/dmlc/dlpack/blob/5c792cef3aee54ad8b7000111c9dc1797f327b59/include/dlpack/dlpack.h#L163-L167

I did a quick check and found that the deleters in TVM and DGL do free `self`, but haven't verified it on PyTorch & Chainer side. Could someone help with this issue? Thanks!

junrushao1994:
I sent a fix on the MXNet side that allows both
https://github.com/apache/incubator-mxnet/pull/15016/

tqchen:
The deleter may not its function argument, but simply implies that the DLManagedTensor no longer want the object. The behavior could be decref if self is also reference counted and shared with others, or delete self if self is a new object and DLManagedTensor is the sole older.

junrushao1994:
@tqchen I agree with your points and suggest we explicitly state that in our protocol.

The bug I introduced in MXNet is that I didn't know the deleter will free the opaque pointer, which causes a double free in DGL.

junrushao1994:
For example, the deleter [here](https://github.com/dmlc/dlpack/pull/38/files#diff-7a886c0f01102a8b14bc674a8c12b553R63) doesn't delete the given handle because the handle is transient.

tqchen:
a PR to clarify the protocol is more than welcomed

junrushao1994:
Will do

Issue #35, opened 2019-03-20
Add Tensor Format option to DLTensor


ryanolson:
As a follow-on to Issue #34, I propose the addition of describing the tensor format for the underlying data in the DLTensor struct. 

This option would consume a single byte with one of two possible values: kDLRowMajor or kDLColumnMajor.

ryanolson:
Upon further thought, one can simple infer based on the strides if the underlying format is row/column major.

Unfortunately, that leaves some ambiguity in the reading of the definition of strides which is allowed to be NULL.

```
  /*!
   * \brief strides of the tensor,
   *  can be NULL, indicating tensor is compact.
   */
  int64_t* strides;
```

https://github.com/dmlc/dlpack/blob/master/include/dlpack/dlpack.h#L140-L144

It is important to be specific about the definition of a NULL value for strides.  I don't have a preference here, but we should explicit state that a NULL value indicates a compact row-major (or column-major) tensor.

tqchen:
It is row-major by default, a PR to clarify things will be more than welcomed

Issue #30, opened 2018-11-02
No mention of unit in the description of `strides`


haowu80s:
```
  /*!
   * \brief strides of the tensor,
   *  can be NULL, indicating tensor is compact.
   */
  int64_t* strides;
```
Does not say if the strides are in bytes or number of elements.I believe `numpy` uses the former and `pyTorch` uses the latter. There does not seem to be an obvious convention.

fdwr:
@haowu80s : I was about to open this issue too because I couldn't tell from the comment (yeah, PyTorch appears to use bytes from the example I saw, and graphics/video formats use bytes when dealing with stride).

PR #40, opened 2019-05-28
Clarification for #30, #35, #39

junrushao1994:
3 things to clarify

1. `DLTensor::strides` means number of elements, not number of bytes. #30 
2. When `DLTensor::strides` is NULL, it is row-majored. #35
3. `DLManagedTensor::deleter` deletes `manager_ctx` as well as its argument `self`. #39


junrushao1994:
CC: @tqchen @zheng-da

PR #42, opened 2019-07-09
adding layout field to store tensor layout (ie. NCHW)

samskalicky:
We need a way to communicate the layout of the tensor (ie. NCHW, NHWC, or NHWC16, etc). This PR includes a char* layout field in the DLTensor.

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -137,6 +137,8 @@ typedef struct {
   DLDataType dtype;
   /*! \brief The shape of the tensor */
   int64_t* shape;
+  /*! \brief The layout of the tensor (ie. NCHW) */
+  char* layout;
   /*!
    * \brief strides of the tensor (in number of elements, not bytes)
    *  can be NULL, indicating tensor is compact and row-majored.


samskalicky:
@junrushao1994

tqchen:
DLPack is a standard exchange format among multiple frameworks(MX, pytorch, chainer etc). We cannot do haste changes without any discussion, and such change will affect ABI and the compatibilities among frameworks.

If there are any usecases or scenarios we have in mind, please open a new  RFC thread. Currently, dlpack assumes row-major layout, the general layout should be the concern of higher-level API.  For example, if we want to pass in data with NCHW4c, the DLpack will simply store a 5 dimensional tensor.

samskalicky:
@tqchen Lets just have the discussion here. How can a framework communicate whether the tensor is NCHW or NHWC? Currently MXNet changes the format depending on the backend, for example MKLDNN prefers NHWC rather than NCHW.

tqchen:
The question boils down to where do we want to encode the layout information. In most deep learning frameworks including MXNet.  The layout is encoded in the computation, for example, conv2d operator in MXNet have a layout argument that explicitly specifies the layout convention.  You can view it as the calling convention of the operator.

Most accelerators or libraries have their preferred variant of layout, which can also be exposed as the properties of the call into the acceleration code.  e.g. MKL-DNN prefers NHWC16c. It is up to the caller to decide whether or do should we do graph rewriting to change the external graph that involves the computation into such a format(for example, already use NHWC16c in previous computation) or explicitly insert a layout conversion.

So to answer your question, frameworks can communicate the layout requirement by declaring these properties when calling external libaries. The tensor in mxnet are just like numpy arrays, which provides the simplicity that allows us to exchange between frameworks.

So what could be the additional benefit of introducing layout as an attribute of a Tensor. Obviously we introduce another level of flexibility, as well as burden to support the operations. Because operators now need to be polymophic to the layout -- you can pass tensor with arbitary layout combinations into an operator, and the operator need to support it. The additional flexibiity is that the function can now decide which layout to return in execution(rather than the planning phase), which may I do not see a strong need for, as most accelerators do have a fixed calling convention(preferred layout). 

In contrary introducing layout to the tensor brings quite a lot of problems. NHWC was a convention commonly used for vision, but what about 3d convolution? naming convention for such layout is quite undefined, that means we cannot agree among frameworks. What will happen when you add an NHWC tensor and NCHW tensor together? Should we raise an error, or should we do direct layout convention. There are a lot more design complexities that comes with this additional complexity.

So to summarize, I think layout information should not be exposed as part of dynamic tensor properties. We can, however, have layout as part of the calling convention(which we already do) and explicitly do smart layout transformation if necessary.

samskalicky:
Thanks @tqchen for the detailed explanation. 

Some motivation here, im working on adding [dynamically loaded custom operator](https://cwiki.apache.org/confluence/display/MXNET/Dynamic+CustomOp+Support) and accelerator support to MXNet. I would like to use DLTensor for both. But I started with a custom tensor format and after discussions with Intel ([see comments at the bottom](https://cwiki.apache.org/confluence/display/MXNET/Bring+your+own+Accelerator)) have added a layout field similar to the proposal here. I would like to see MKLDNN disentangled from MXNet so having this support is critical to Intel adopting these interfaces in MXNet.

One could make the same argument about the context, why dont we just pass that when calling external libraries as opposed to encoding it into the tensor? 

>  Because operators now need to be polymophic to the layout -- you can pass tensor with arbitary layout combinations into an operator, and the operator need to support it.

I would say that they dont _need_ to support, and could error out with an appropriate message as a starting point. If a framework wanted to do an automagic conversion from NCHW to NHWC, or had the described polymorphic capabilities then that would be enabled by tagging the tensor with the layout. 

By not specifying a specific layout, we're implicitly defining an assumption of some layout (maybe NCHW). Or at least requiring an orthogonal channel for this info to be passed (as you describe).

tqchen:
Again, the layout can be defined implied in the calling convention(as the interface spec) and does not have to be part of the tensor attribute, so it is not something that have to be part of DLTensor.

While adding layout as a field might be a way to solve your problem technically, it may not be the only way, nor may it be the best way. 

For the case of MKLDNN, because it assumes a fixed internal layout, you can simply query the interface about the layout requirement, and doing inference to propagate that through or insert a conversion. Internally, make it a special context and you don’t need to worry about layout. 

We can talk about why context is encoded. This is because things can indeed be polymorphic with respect to the context, for most devices, they can support any shape and op. 

Layout, on the other hand, is rather tied to the hardware, its decision should be best made at planning or optimization phase, rather than the execution phase, which makes it unnecessary to be carried here. I would think twice about introducing it as a property of a runtime object.

Every new feature we introduce also will introduce technical debt. DLPack is an exchange standard, which means every data structure change need to be agreed and adopted by all parties of interest. It is easier to get everyone agree on what set of context to support, but it is much harder to come up with a standard set of layout and how to store them. At least the NHWC string was rather limited and could not serve as a standard due to its strong ties to vision but not the rest of the cases

samskalicky:
Some references of different layouts for posterity
https://github.com/mratsim/Arraymancer/issues/132
https://github.com/pytorch/pytorch/issues/1220#issuecomment-477460538

tqchen:
Sure. Named axis dimension, if people can agree on the convention(which I don't know yet what it should be), may or may not be a useful feature. 

The reason that MKL-DNN requires this layout field is mainly due to the fact that it consolidates the planning phase with the execution phase. So each operation can greedily choose what is the best output layout depending on the input layout. 

This approach(runtime greedy per op used by MKL-DNN) however, is not necessarily optimal globally. A better approach would be to have a planning phase that decides on the layout, and then execute each operator. In the latter case, the planning phase needs to have layout, but runtime does not.

Given that most of the Tensors/Arrays in MXNet/Pytorch/TF are still un-named. It will break the backward compatibility of the current tensors who are backed by DLTensor. For example, if you take an arbitrary ndarray in mxnet, you do not know its layout from the data structure itself.  There are certain values in such un-named arrays due to its simplicity.

Named version of array, or layout processing, could be built on top of this un-named versions. Right now most frameworks use calling the convention to do so. This does not necessarily mean we have to fix to a layout for exchange, instead, you can also have the API declare its requirements and ability in terms of layout it supports and prefer and mix and match from there.

PR #44, opened 2019-12-16
add bfloat data type support

ElaineBao:
This is to add a new data type (bfloat), so that it can support mxnet bfloat16 development.

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -80,6 +80,7 @@ typedef enum {
   kDLInt = 0U,
   kDLUInt = 1U,
   kDLFloat = 2U,
+  kDLBfloat = 3U,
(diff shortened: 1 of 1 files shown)


ElaineBao:
@tqchen Can you please help to review this change? Thank you!

tqchen:
bfloat is certainly something that could potentially be added to DLPack. 

However, given that this creates a major change to the current DLPack, it should be done in a form of an RFC, getting interested party involved, instead of directly bring the type code in as it is. 

Alternatively, given that the support is not entirely fixed, we can allow projects feel free to add the additional type code in those sub-projects, before bringing the type forward to DLPack

pengzhao-intel:
@tqchen the RFC is sent out. 
Could we re-open the PR in case others want to take a review?

PR #47, opened 2019-12-22
fix two typos.

csukuangfj:
Fix two typos in the documentation.


szha:
Thanks for the patch.

PR #58, opened 2021-01-21
Support complex numbers in DLPack

### LINKED ISSUE #50: Support complex numbers in DLPack protocol
The DLPack protocol doesn't seem to have a representation of complex numbers. I don't know whether this is intentional (e.g., is the intent that clients use a structure-of-arrays representation of the real and imaginary parts?) or whether it is a missing feature.

(I therefore chose to leave this case unimplemented when adding DLPack support to JAX, but it would be nice to add it once the protocol is fixed.)

leofang:
Close #50. 

This PR adds the last missing piece IMHO to DLPack, namely to support complex numbers commonly used in scientific computing and also increasingly in certain machine learning tasks.

As discussed in #50, there seems to be a need to support two different kinds of formats, namely structs of arrays (SoA) and arrays of structs (AoS). The latter (AoS) is the standard choice in C/C++/CUDA/HIP/Python/Numpy/CuPy/etc, so I took the liberty of simply calling it `kDLComplex`. We haven't converged on naming in #50, though, and I'm open to suggestions 🙂

**QUESTION**: Should I bump the DLPack version in this PR, or we can wait until the actual release?

cc: @rgommers @tqchen @hawkinsp

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -81,8 +81,7 @@ typedef enum {
   kDLUInt = 1U,
   kDLFloat = 2U,
   kDLBfloat = 4U,
-  kDLComplex = 5U,    // C/C++/Python layout (compact struct per complex number)
-  kDLComplexS = 6U,  // alternative layout (a struct with two pointers to real & imag arrays)
+  kDLComplex = 5U,  // C/C++/Python layout (compact struct per complex number)
 } DLDataTypeCode;
 
 /*!
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -90,7 +90,11 @@ typedef enum {
   kDLOpaqueHandle = 3U,
   /*! \brief bfloat16 */
   kDLBfloat = 4U,
-  kDLComplex = 5U,  // C/C++/Python layout (compact struct per complex number)
+  /*!
+   * \brief complex number
+   * C/C++/Python layout: compact struct per complex number)
+   */
+  kDLComplex = 5U,
 } DLDataTypeCode;
 
 /*!
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -81,6 +81,8 @@ typedef enum {
   kDLUInt = 1U,
   kDLFloat = 2U,
   kDLBfloat = 4U,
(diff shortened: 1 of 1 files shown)


include/dlpack/dlpack.h
```diff
@@ -81,6 +81,8 @@ typedef enum {
   kDLUInt = 1U,
   kDLFloat = 2U,
   kDLBfloat = 4U,
+  kDLComplex = 8U,    // C/C++/Python layout (compact struct per complex number)
```
tqchen, line 84: do do not need to assign bitmasking. the number 3U was reserved for opaque handle, we can directly assign 5 here

tqchen (changes requested):

include/dlpack/dlpack.h
leofang: OK, changed in b6a64fa. @tqchen What's it for? I don't find it documented anywhere.

include/dlpack/dlpack.h
```diff
@@ -81,8 +81,8 @@ typedef enum {
   kDLUInt = 1U,
   kDLFloat = 2U,
   kDLBfloat = 4U,
-  kDLComplex = 8U,    // C/C++/Python layout (compact struct per complex number)
-  kDLComplexS = 16U,  // alternative layout (a struct with two pointers to real & imag arrays)
+  kDLComplex = 5U,    // C/C++/Python layout (compact struct per complex number)
+  kDLComplexS = 6U,  // alternative layout (a struct with two pointers to real & imag arrays)
```
tqchen, line 85: Given that the Struct of Array is less conventionally defined(we need to define how the layout is, right now seems only well defined for compact array), let us first start with Complex(which is easier to agree on) then discuss the SOA case later.

leofang: Sounds good, @tqchen. This is also what I suggested in #50. I removed `kDLComplexS` in 40beada. PTAL.

leofang:
btw @tqchen Should I bump the DLPack version in this PR?

leofang:
@tqchen I noticed there appeared a conflict with the main branch. I just fixed it.

tqchen (approved):

leofang:
Thanks, @tqchen! I still need to know when the DLPack version will be bumped, though.

Issue #61, opened 2021-02-09
[RFC] Rename DLContext to DLDevice


icemelon9:
This RFC proposes to rename the `DLContext` to `DLDevice`. `DLContext` indicates a device for Tensor and ops. Two main reasons for this change:

- The "DLContext" name doesn’t intuitively reflect its semantics of a device.
- Mainstream frameworks including [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/device) and [PyTorch](https://pytorch.org/docs/stable/tensor_attributes.html?highlight=device#torch.torch.device) both use `device` to represent the device to run. I think it’d be good that we use similar terminology to reduce the confusion.

This change will require downstream projects to change accordingly. Feel free to bring up your thoughts and concerns.

Meanwhile, we have a similar [RFC](https://discuss.tvm.apache.org/t/rfc-rename-tvmcontext-to-tvmdevice/9090) for the name change in TVM discuss forum.

junrushao1994:
I agree with the name change. `Context` doesn't really mean anything - its meaning varies depending on its context...

However, to keep people aligned well, I think we need to take a gradual approach:
1) Step 1: provide `DLDevice` as an alias type of `DLContext` by `typedef`
2) Step 2: in the next release, exchange `DLDevice` as the name of the struct, and `DLContext` as an alias
3) Step 3: in the next release, emit a deprecation warning if anyone uses `DLContext` via `__attribute__((deprecated)) `
4) Step 4: remove `DLContext` completely.

szha:
While I agree that "context" is often overloaded, the name "device" can also be confusing as sometimes a physical device can be partitioned into smaller units for execution. For example, the MIG support on A100 allows partitioning a GPU into MIG instances that are all backed by the same GPU device.

junrushao1994:
Yeah I agree with @szha...so what will be the best name then?

tqchen:
Given that Pytorch and TensorFlow adopts the term Device I think it might be a good chance for converging from the API compatibility point of view.

Since naming change does not change the underlying ABI layout, change to `DLDevice` and provide type aliasing might be fine. It would be great to create a quick test case if `from_dlpack` function still works  once we upgraded to the new name, while importing from DLPack exported by  say pytorch(which uses the old name). My guess is that unless there is a symbolc conflict, things should work out of box.

tqchen:
cc @kkraus14 @leofang @prigoyal @zdevito @rgommers @oleksandr-pavlyk

[...]

PR #62, opened 2021-02-19
Rename DLContext to DLDevice

icemelon9:
This is the PR for the RFC #61 .

As discussed in the RFC, we will rename `DLContext` to `DLDevice` and provide `DLContext` as an alias via typedef.

cc: @tqchen @junrushao1994 @leofang @szha @kkraus14


include/dlpack/dlpack.h
```diff
@@ -64,14 +64,16 @@ typedef enum {
...
  */
 typedef struct {
   /*! \brief The device type used in the device. */
   DLDeviceType device_type;
   /*! \brief The device index */
   int device_id;
-} DLContext;
+} DLDevice;
+
+typedef DLDevice DLContext;
```
tqchen, line 76: Please add a comment about the alias and indicating that we will consider remove it in the next release

tqchen (changes requested):

include/dlpack/dlpack.h
icemelon9: added the comment.

tqchen (approved):

junrushao1994 (approved):

tqchen:
Thanks @icemelon9 ! This PR is being merged

leofang:
Thanks!

include/dlpack/dlpack.h
```diff
@@ -144,8 +149,8 @@ typedef struct {
    * \endcode
    */
   void* data;
-  /*! \brief The device context of the tensor */
-  DLContext ctx;
+  /*! \brief The device of the tensor */
+  DLDevice device;
```
leofang, line 153: @tqchen @icemelon9 Unfortunately this breaks existing C code. Could you please revert the field name to `ctx`?

tqchen: @leofang this is intentional to keep the code consistent. The existing codebase can either stay at previous version, or upgrade the code to move to the new version. Exchange won't be affected by the version being used

leofang: Ah OK, thanks. I failed to see this implication earlier.

PR #69, opened 2021-05-11
[VERSION] Bump to v0.5

tqchen:
Prepares for v0.5 release

- Remove the deprecated DLContext
- Updated the comment on kDLCUDAHost to remove remark
  about bit-masking as we cannot introduce future
  enums that depends on this behavior.

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -13,7 +13,7 @@
 #endif
 
 /*! \brief The current version of dlpack */
-#define DLPACK_VERSION 040
+#define DLPACK_VERSION 050
 
 /*! \brief DLPACK_DLL prefix for windows */
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


tqchen:
cc @kkraus14 @leofang @prigoyal @zdevito @rgommers @oleksandr-pavlyk @leofang @junrushao1994 @icemelon9

leofang:
@tqchen If you'd like me to put together a separate RFC for managed memory etc, can we please hold off the new release? I can work on it later this week but not today.

tqchen:
Thanks @leofang I believe the managed(unified) memory could use a bit more thoughts mainly due to the current state of adoption of such memory types in common frameworks and might worth push to another cycle.

kkraus14:
I'm working on getting the `dlpack` conda-forge packages updated since it was still on 0.3.

Any chance someone can tag the release of 0.5 for the conda-forge auto-upgrade bot to kick in? Thanks!

tqchen:
just did that https://github.com/dmlc/dlpack/releases/tag/v0.5

rgommers:
Out of interest, why is there a `dlpack` package in conda-forge? I thought the idea was for every library to vendor `dlpack.h` - when would one want to depend on the package instead?

kkraus14:
> Out of interest, why is there a `dlpack` package in conda-forge? I thought the idea was for every library to vendor `dlpack.h` - when would one want to depend on the package instead?

For C/C++ libraries you can often leak the dlpack types in your public interfaces, in which case a library building against your library also needs the dlpack headers. I.E. we have this for `libcudf`

PR #68, opened 2021-05-11
Rename kDLGPU to kDLCUDA, and kDLCPUPinned to kDLCUDAHost

YuchenJin:
This is the PR for the RFC #67.

As discussed, we do the following renaming:
`kDLGPU` -> `kDLCUDA`
`kDLCPUPinned` -> `kDLCUDAHost`

Issue #51, opened 2020-01-31
Specify Python embedding of DLPack tensors


hawkinsp:
DLPack only specifies a C++ API, but in practice there's a Python embedding that multiple frameworks support (via Python capsules) that does not seem to be formally specified or standardized.

The protocol seems to be:
* producers embed a `DLPackManagedTensor` as Python capsule with name `"dltensor"`.
* when a consumer consumes a `DLPackManagedTensor`, it renames the capsule to `"used_dltensor"` so the same capsule cannot be consumed twice.
* different frameworks seem to act differently as to how a consumer should treat a capsule destructor. MXNet seems to remove the capsule destructor, but PyTorch seems to leave it alone (I may have misread the code in either of these two cases.) It would be good to clarify what the correct behavior is. For JAX, I chose to remove the capsule destructor.

tqchen:
Thanks @hawkinsp !The protocol is correct. In terms of removal of destructor. PyTorch's destructor checkes the used_dltensor flag before destruction, so the effect is same as destructor removal.

A PR is more than welcomed.

hawkinsp:
There's still an issue I think. Consider the following case:
* Hypothetical framework F provides a capsule with a destructor that does not check the capsule name.
* PyTorch consumes a DLPack, setting the capsule name to `used_dltensor` but leaving the capsule destructor alone.
* You have a potential double free: the tensor may be destroyed once when F's capsule destructor is called, and once when PyTorch's destroys the DL tensor via the callback in `DLManagedTensor`.

So you need at least one of the following rules:
a) frameworks must remove the capsule destructor when consuming a DLPack, or
b) frameworks must check the capsule name matches "dltensor" in their capsule destructors, and otherwise do nothing.

It would probably be good to specify which.

tqchen:
you are right, I think it is probably a good way to specify the deletion of the destructor. To keep behavior compact to the pytorch one, we can recommend the destructor to check the name as well

vadimkantorov:
Also it would be good to have somewhat "recommended" complete ctypes wrapper in the repo. Currently it is partly in the NumPy example.

I made my own in https://github.com/vadimkantorov/pydlpack/blob/master/dlpack.py

tqchen:
This is now part of https://data-apis.org/array-api/latest/design_topics/data_interchange.html

eric-wieser:
The linked page does not specify anything about what the capsule destructor should do. Would it make sense to specify an exact implementation?
```C
[...]

PR #71, opened 2021-05-19
Add `kDLROCMHost` and `kDLCUDAManaged`

### LINKED ISSUE #70: [RFC] Add `kDLROCMHost` and `kDLCUDAManaged`
From https://github.com/dmlc/dlpack/issues/67#issuecomment-834784656 and https://github.com/dmlc/dlpack/issues/67#issuecomment-835669619: 

I am proposing to add two new device types, following the line of #67:
1. `kDLROCMHost`
2. `kDLCUDAManaged`

The first addition is to mirror the current relation (since v0.5) between CUDA and ROCm now that we have `kDLCUDA`, `kDLCUDAHost`, and `kDLROCM`. ROCm also provides pinned/page-locked memory, so this is legit.

The second addition is for CUDA managed/unified memory, which does not belong to either host or device but to both. It seems natural to me to have a standalone type for it. ROCm currently does not provide managed memory, so we could add it in the future once AMD implements it.

Both additions seem straightforward for me to add without any issue, as they are orthogonal to existing device types (as they should be).

cc: @rgommers @tqchen @jakirkham @kkraus14 @kmaehashi @emcastillo @asi1024

leofang:
Close #70.

btw I feel `kDLExtDev` should possess a very high value (like 999) so that we don't get clogged in the `DLDeviceType` enum spectrum, but changing its value is something beyond the scope of this PR.

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -55,11 +55,19 @@ typedef enum {
   /*! \brief ROCm GPUs for AMD GPUs */
   kDLROCM = 10,
   /*!
+   * \brief Pinned ROCm CPU memory allocated by hipMallocHost
+   */
+  kDLROCMHost = 11,
+  /*!
    * \brief Reserved extension device type,
    * used for quickly test extension device
    * The semantics can differ depending on the implementation.
    */
   kDLExtDev = 12,
+  /*!
+   * \brief CUDA managed/unified memory allocated by cudaMallocManaged
+   */
+  kDLCUDAManaged = 13,
 } DLDeviceType;
 
 /*!
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -76,7 +76,10 @@ typedef enum {
 typedef struct {
   /*! \brief The device type used in the device. */
   DLDeviceType device_type;
-  /*! \brief The device index */
+  /*!
+   * \brief The device index.
+   * For vanilla CPU memory, pinned memory, or managed memory, this is set to 0.
+   */
   int device_id;
 } DLDevice;
 


tqchen (approved):

leofang:
Thanks, @tqchen!

seberg:
Since it requires checking 4 devices now (and more might be added?), should there be a macro or other scheme to check whether memory is transparently accessible from the CPU?

leofang:
@seberg Sounds like you want to add a query function to the DLPack header so that everyone can reuse it? I think it is a good idea given that you also brought up [another common API on deleter](https://github.com/dmlc/dlpack/issues/51#issuecomment-849794854).

leofang:
@tqchen and all: Could you make a new DLPack release (v0.6) to include this change please? Thanks!

jakirkham:
Should we doing something about Vadim's concern first? Asking as it may be harder to change after a release :)

leofang:
@jakirkham That'd break ABI and I don't think the recent DLPack iterations have done that. I'd suggest for @vadimkantorov to open an issue to discuss that point first, likely we also need to handle #74 so that we can signal ABI breaking. 

I am requesting a new release simply based on the experience with the previous release #69: it was done almost right after #68 was merged, as the ABI is preserved. I expect the same for this PR.

tqchen:
Changing to int to int32_t likely won't break the ABI and i think it can be a welcomed change

vadimkantorov:
I also read online that enums with values < 255 must occupy 1 byte - could be good to have it stated somewhere in the docs. Unfortunately, it seems that in C one cannot enforce bit-width for enum type.

leofang:
@tqchen Friendly nudge for a v0.6 release. I'd like to make this PR available asap. Thanks!

PR #72, opened 2021-05-25
Add ABI version.

### LINKED ISSUE #34: Future ABI compatibility
DLPack’s vision of a common in-memory tensor format that spans device and memory types is fantastic. 

However, in its form, there is no upgrade path for adding new items to the either the DLTensor or DLManagedTensor structs in way that would maintain ABI compatibility. 

I would like to propose the addition of two components to the DLTensor struct. This will break current ABI compatibility, but will in the long run future proof the design. 

Additions to DLTensor
1) `unt8/16_t version`
2) `uint64_t future_bytes`

Adding the version allows the receiving library to determine if the DLTensor can be consumed.  The receiver may not have a matching version, but as long as it knows the version it can make a decision on if the data can be correctly used. 

Adding `future_bytes` allows for the addition of new options to DLTensor. One of which might be data layout, ie row-major or column major (c-format vs FORTRAN).  I will open a separate issue for this feature.

hameerabbasi:
Things to be discussed:

- Do we put the `DLVersionInfo` inside `DLTensor`? If so, this is an ABI breaking change in itself.
- Is the naming convention okay?
- Do we even need a function?

Fixes #34.

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -12,8 +12,15 @@
 #define DLPACK_EXTERN_C
 #endif
 
-/*! \brief The current version of dlpack */
-#define DLPACK_VERSION 050
+/*! \brief The current version of dlpack
+ *  This is stored in the form `major | (minor << 8) | (bugfix << 16)`
+ */
+#define DLPACK_VERSION 0 | (5 << 8) | (0 << 16)
+
+/*! \brief The ABI version of dlpack
+ *  This is used to indicate backwards-incompatible ABI changes.
+ */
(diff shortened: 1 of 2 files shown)


include/dlpack/dlpack.h
```diff
@@ -12,8 +12,15 @@
 #define DLPACK_EXTERN_C
 #endif
 
-/*! \brief The current version of dlpack */
-#define DLPACK_VERSION 050
```
hameerabbasi, line 16: Note here, the leading `0` meant this was being stored as an octal value, so it was decimal `40`.

PR #78, opened 2021-09-27
Add oneapi support

oleksandr-pavlyk:
This PR proposes to extend `DLDeviceType` enum with 3 new entries: 
`kDLONEAPI_GPU`, `kDLONEAPI_CPU`, and `kDLONEAPI_ACCELERATOR`.

This adds DLPack support for OneAPI SYCL root-devices, addressable with 
[filter-selector](https://github.com/intel/llvm/tree/sycl/sycl/doc/extensions/FilterSelector), e.g. ``sycl::ext::oneapi::filter_selector("gpu:device_id")`` for 
`kDLONEAPI_GPU` devices.

Two parties wishing to zero-copy exchange [USM allocations](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_usm_allocations) using DLPack need to bind their
allocations to the [default platform context](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/PlatformContext/PlatformContext.adoc) (implicitly used by ``sycl::queue(dev)``
constructor), ensuring that both parties use the same `sycl::context` associated with the agreed
upon SYCL device, thus ensuring that USM allocations made by one party are accessible
to another.

An application `apps/from_usm_ndarray` is included in this PR to demonstrate working prototype, 
compiled with DPC++ 2021.3, or Open Source LLVM-Sycl compiler [release 2021-07](https://github.com/intel/llvm/releases/tag/2021-07).


oleksandr-pavlyk:
@tqchen @leofang @rgommers Please provide feedback

include/dlpack/dlpack.h
```diff
@@ -68,6 +68,9 @@ typedef enum {
    * \brief CUDA managed/unified memory allocated by cudaMallocManaged
    */
   kDLCUDAManaged = 13,
+  kDLONEAPI_GPU = 14,
+  kDLONEAPI_CPU = 15,
+  kDLONEAPI_ACCELERATOR = 16,
```
leofang, line 73: Could you add short comments for each entry for better clarity?

leofang (commented):
Thanks, @oleksandr-pavlyk. One very high level (and perhaps naive, since I am not familiar with oneAPI) question: My understanding is oneAPI can target non-Intel devices, such as NVIDIA or AMD GPUs. What happens, then, when a oneAPI-based library targeting CUDA/HIP/OpenCL exchanges with another library implemented natively on one of these devices? 

Say I have `kDLONEAPI_GPU` --> `kDLCUDA`. Does an importer handle this in their DLPack implementation? Or is there a runtime flag somewhere that we can check and see if two libraries are actually running on an NVIDIA GPU? Or do we only allow exchanges between two oneAPI libraries (maybe context sharing is not trivial in oneAPI)?

oleksandr-pavlyk:
> Thanks, @oleksandr-pavlyk. One very high level (and perhaps naive, since I am not familiar with oneAPI) question: My understanding is oneAPI can target non-Intel devices, such as NVIDIA or AMD GPUs. What happens, then, when a oneAPI-based library targeting CUDA/HIP/OpenCL exchanges with another library implemented natively on one of these devices?
> 
> Say I have `kDLONEAPI_GPU` --> `kDLCUDA`. Does an importer handle this in their DLPack implementation? Or is there a runtime flag somewhere that we can check and see if two libraries are actually running on an NVIDIA GPU? Or do we only allow exchanges between two oneAPI libraries (maybe context sharing is not trivial in oneAPI)?

I do mean to try this scenario out, hopefully later this week, so for now I will be speaking hypothetically. Implementation of SYCL API is provided by a [backend](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_the_sycl_backend_model). Open source LLVM SYCL compiler has such a [CUDA backend](https://github.com/intel/llvm/blob/sycl/sycl/doc/GetStartedGuide.md#build-dpc-toolchain-with-support-for-nvidia-cuda).

In principle, SYCL entities (such as SYCL device, SYCL context) allow one to retrieve the corresponding native object stored by the backend, using [backend interoperability](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:backend-interoperability). This means, however, that for an importer, say cupy, to be able to import CUDA memory exported with `device_type=kDLONEAPI_GPU`, one needs oneAPI runtime to check if the backend is `sycl::backend::cuda`, and to get appropriate CUDART objects.

One would also need oneAPI runtime to [check the type of USM allocation](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_unified_shared_memory_pointer_queries), since DLPack uses different device codes for USM-device allocation (`kDLGPU`) vs.  USM-shared allocation (I surmise this should correspond to `kDLCUDAManaged`)

include/dlpack/dlpack.h
oleksandr-pavlyk: Addressed.

apps/from_usm_ndarray/usm_ndarray_dlpack.cpp
```diff
@@ -0,0 +1,99 @@
...
+	break;
+    default:
+	throw std::runtime_error("Can not handler these types");
+    }
+
+    sycl::queue q(*d_ptr);   // queue uses default context per root device
+    sycl::context ctxt = q.get_context();
+    delete d_ptr;
+
+    sycl::usm::alloc kind = sycl::get_pointer_type(tensor.data, ctxt);
```
csullivan, line 68: In a zero-copy exchange of a USM allocation between two applications, is there additional handling that generated code needs to employ when the allocation kind is `shared` vs `device`? Said differently, if a backend consumes a dltensor with  kDLONEAPI_GPU device type, will it be sufficient for the runtime to consume allocations of these two kinds interchangeably?

apps/from_usm_ndarray/usm_ndarray_dlpack.cpp
```diff
@@ -0,0 +1,99 @@
...
+    switch (kind) {
+    case sycl::usm::alloc::shared:
+	std::cout << "USM-shared allocation-based array" << std::endl;
+	break;
+    case sycl::usm::alloc::device:
+	std::cout << "USM-device allocation-based array" << std::endl;
+	break;
+    case sycl::usm::alloc::host:
+	std::cout << "USM-host allocation-based array" << std::endl;
+	break;
```
csullivan, line 78: Can an allocation made outside of `sycl::usm`, e.g. by numpy, be zero copied into a DLTensor of type `kDLONEAPI_CPU` for use in a sycl program? 

I wonder the same about CUDA/OpenCL buffer allocations, but limit the question to CPU allocations for simplicity.

apps/from_usm_ndarray/usm_ndarray_dlpack.cpp
oleksandr-pavlyk: Once the importer figures out the SYCL context associated with USM allocation, any SYCL kernel submitted to a queue associated with such a context can access USM memory irrespective of the USM allocation type.  The runtime takes care of that.

Once USM allocation is made, there is no way to change its allocation type. The type of USM allocation is queried from the pointer and the sycl context it is associated with.  If query is done against a different SYCL context, the returned type is "unknown"

In this particular application, line 68 serves to demonstrate that query worked correctly, and thus correct context was reconstructed.

apps/from_usm_ndarray/main.py
```diff
@@ -0,0 +1,152 @@
...
+    # For 0-dim ndarrays, strides and shape will be NULL
+    dl_tensor.shape = vec_as(array.shape, ctypes.c_int64)
+    dl_tensor.strides = vec_as(array.strides, ctypes.c_int64)
+    dl_tensor.byte_offset = (array._pointer - array.usm_data._pointer)
+    return dl_tensor
+
+
+def main():
+    usm_ary = dpt.usm_ndarray((3,1,30), dtype="f4", buffer="device");
+    usm_ary[:] = np.random.rand(3, 1, 30)
```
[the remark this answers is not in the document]
oleksandr-pavlyk: One can cast on the host, if it makes the code clearer, but it is up to Python code in ``__setitem__`` to do the right thing, such as copy while coercing in a kernel executed on the device associated with USM allocation.

apps/from_usm_ndarray/usm_ndarray_dlpack.cpp
oleksandr-pavlyk: Yes. Host memory can be access from within SYCL kernels through use of ``sycl::buffer``.  

For example, see https://github.com/IntelPython/dpctl/blob/master/examples/pybind11/use_dpctl_syclqueue/pybind11_example.cpp#L61

The function `offloaded_array_mod` works on a regular NumPy array allocated using NumPy's host allocator:

https://github.com/IntelPython/dpctl/blob/master/examples/pybind11/use_dpctl_syclqueue/example.py#L39

csullivan:
I see that the device, and therefore the device aspect, (type: cpu/gpu/accelerator) can be [queried](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#_unified_shared_memory_pointer_queries) from a USM allocation. Could it be sufficient to then only introduce `kDLOneAPI`? 

I started wondering along this track based on your above comment about kDLCPU as it does seem strange to have both it and kDLOneAPI_CPU. If the only distinction comes down to whether data was allocated via a usm_allocator, then perhaps a single entry to the DLDeviceType could suffice.

csullivan:
@oleksandr-pavlyk Thank you for the detailed follow up on this. Given the flexibility to derive the sycl::device from the device id as you've shown, my preference would be to introduce the single DLDeviceType::kDLOneAPI as the initial OneAPI support in DLPack.

PR #79, opened 2021-10-11
[Apps] Replace the from_numpy demo with a more comprehensive example.

jwfromm:
@junrushao1994 and I put together an end-to-end example showing how copy free transfers between two DLPack compatible frameworks can be performed. This is a strictly more functional demo of the capabilities of DLPack than the previous `from_numpy` example and I think will be a solid base to build other integrations on top of.


jwfromm:
@tqchen Can you take a quick look at this and let us know what you think?

jwfromm:
Ah I hadn't seen your proposal in that thread. I like it a lot. Let me refactor this PR into that proposal. One open question is if we want to run any tests on the python library since it seems like this repo isnt set up for that.

tqchen:
superseded

PR #84, opened 2021-11-11
[Apps] Replace the from_numpy demo with a more comprehensive example: with improvements

oleksandr-pavlyk:
This is pull request builds of good work done in #79 but addresses several issues raised by myself during review.

The issues are:

   1. `restype` of `ctypes.python_api.PyCapsule_New` is set to ensure that the returned type is a Python object of type PyCapsule, rather than a Python object of type `int`.
   2. The `DLManagedTensor` class no longer defines `__array_interface__` property, but reconstructs the array from capsule by calling a function defined in `to_numpy`.
   3. Added a test-runner file which does not assume that TVM is installed (tests based on NumPy only).

An attempt to consider this changes to the original repo from which request #79 pulls went unanswered.

Hence, I am opening this PR in parallel to #79 as an alternative.

@junrushao1994


apps/numpy_dlpack/dlpack/dlpack.py
```diff
@@ -0,0 +1,102 @@
+import ctypes
+
+_c_str_dltensor = b"dltensor"
+
+
+class DLDeviceType(ctypes.c_int):
```
junrushao1994, line 6: Shall we also add docstrings to these classes correspondingly?

apps/numpy_dlpack/dlpack/to_numpy.py
```diff
@@ -0,0 +1,77 @@
...
+import numpy as np
+from .dlpack import _c_str_dltensor, DLManagedTensor, DLTensor
+
+ctypes.pythonapi.PyCapsule_IsValid.restype = ctypes.c_int
+ctypes.pythonapi.PyCapsule_IsValid.argtypes = [ctypes.py_object, ctypes.c_char_p]
+
+ctypes.pythonapi.PyCapsule_GetPointer.restype = ctypes.c_void_p
+ctypes.pythonapi.PyCapsule_GetPointer.argtypes = [ctypes.py_object, ctypes.c_char_p]
+
+def array_interface_from_dl_tensor(dlt):
```
junrushao1994, line 11: Shall we prefix this with underscore?

```suggestion
def _array_interface_from_dl_tensor(dlt):
```

junrushao1994:
Overall it looks good to me! Also CC: @jwfromm @tqchen

apps/numpy_dlpack/dlpack/dlpack.py
oleksandr-pavlyk: Added. Please provide feedback.

apps/numpy_dlpack/dlpack/to_numpy.py
oleksandr-pavlyk: Renamed: both the definition and the invocation.

junrushao1994 (approved):
The PR looks in a good shape. Thanks @oleksandr-pavlyk!

tqchen (approved):

PR #86, opened 2021-11-12
[ENUM] Declare DLDeviceType as int32  enum in c++

tqchen:


tqchen:
Helps to further clarify the specification

include/dlpack/dlpack.h
```diff
@@ -26,6 +26,12 @@
 #define DLPACK_DLL
 #endif
 
+#ifdef __cplusplus
+#define DLPACK_INT32_ENUM_TYPE : int32_t
```
oleksandr-pavlyk, line 30: Perhaps `#include <cstdint>` before using `int32_t`.

oleksandr-pavlyk (approved):
Took me a minute to parse the change. Compiled with C++ compiler to be sure too.

include/dlpack/dlpack.h
```diff
@@ -172,7 +178,7 @@ typedef struct {
   /*! \brief The device of the tensor */
   DLDevice device;
   /*! \brief Number of dimensions */
-  int ndim;
+  int32_t ndim;
```
csullivan, line 181: Could consider using `DLPACK_INT32_ENUM_TYPE` here to avoid future divergence.

csullivan: I believe we want to update [Lines 90/96](https://github.com/dmlc/dlpack/pull/86/files#diff-1564e5abd0da5cf46e82b28b0058996df944007494520a2b0a6b629880eaa097R96) as well.

`int device_id` -> `int32_t device_id` or equivalently 
-> `DLPACK_INT32_ENUM_TYPE device_id` as mentioned above.

csullivan (changes requested):

include/dlpack/dlpack.h
tqchen: in this case this is not an enum, mainly to clarify. `DLPACK_INT32_ENUM_TYPE` is indeed confusing, going to change `DLPACK_INT32_ENUM_ANNOTATION`

csullivan: In that case, I think the only change we would want is `int device_id -> int32_t device_id`.

csullivan (approved):

Issue #97, opened 2022-02-28
Endianness support is missing


leofang:
Complaints on endianness has been something I've recurrently seen (ex: CuPy https://github.com/cupy/cupy/issues/3652 and mpi4py https://github.com/mpi4py/mpi4py/issues/177), and I anticipate at some point we'd start receiving bug reports on this. Apparently there is at least a few communities out there (astropy and hdf5) that prefer (or could work with) non-native (that is, big) endianness data. This causes problems if two libraries exchange but do not communicate the endianness for how to interpret the data.

I suggest two possible solutions:
1. Add an endianness enum (big, little, native, etc) and include it in `DLDataType` as a new struct member:
   - Cons: This will be an API/ABI incompatible change, unfortunately.
2. Alternatively, we could apply a bitwise mask to `DLDataType::code` to make it carry this information:
   - The mask should be designed such that when not applied it refers to the little endianness, the de facto standard used by all projects so far

cc: @tqchen @rgommers @tirthasheshpatel

rgommers:
FITS as an astrophysics-specific big-endian format is annoying indeed. Your `mpi4py` link also seems to come from that community. I can't think of many other fields where this comes up, but it's hard to be sure.

A few thoughts:
- Ignoring it completely yields correctness issues; transparently converting to native endianness in I/O routines as CuPy just did makes sense. 
- Right now, DLPack implementations should check for this and raise an exception on the exporter side. Not doing so looks like a pretty severe bug to me. 
- Solution 1 above: there's a longer list of info we'd like to see captures (like readonly), when that is done this could be taken along - it's not an extra/separate ABI break.
- Solution 2 above: not a fan of such an obscure solution.
- It's not clear to me that either solution will be generally helpful. If the vast majority of libraries supporting DLPack do not support non-native byteorder, then an exception will still be raised on the importer side once endianness info is present. Hence not much change compared to the current state where the exporter must raise. So it's probably okay to add if it can be taken along in a larger set of changes, but is not worth a separate (breaking or convoluted) change.

tirthasheshpatel:
> * It's not clear to me that either solution will be generally helpful. If the vast majority of libraries supporting DLPack do not support non-native byteorder, then an exception will still be raised on the importer side once endianness info is present. Hence not much change compared to the current state where the exporter must raise. So it's probably okay to add if it can be taken along in a larger set of changes, but is not worth a separate (breaking or convoluted) change

[...]

PR #98, opened 2022-03-08
Add a note about non-native endianness

tirthasheshpatel:
Addresses #97 

Added a note saying non-native endianness isn't supported and that exporting such arrays must raise an error. @tqchen Do we want this on the roadmap or should we consider it out of scope?

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -126,7 +126,8 @@ typedef enum {
 } DLDataTypeCode;
 
 /*!
- * \brief The data type the tensor can hold.
+ * \brief The data type the tensor can hold. Currently, non-native endianness isn't
+ * supported. Attempt to export such array must raise an error.
  *
  *  Examples
  *   - float: type_code = 2, bits = 32, lanes=1
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -126,8 +126,9 @@ typedef enum {
 } DLDataTypeCode;
 
 /*!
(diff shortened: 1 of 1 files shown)


edgarriba (approved):

tqchen (approved):

include/dlpack/dlpack.h
```diff
@@ -126,7 +126,8 @@ typedef enum {
 } DLDataTypeCode;
 
 /*!
- * \brief The data type the tensor can hold.
+ * \brief The data type the tensor can hold. Currently, non-native endianness isn't
```
tqchen, line 129: How about change to data type is assumed to follow the native endian-ness. An explicit error message should be raised when attempting to export an array with non-native endianness

tqchen (changes requested):

include/dlpack/dlpack.h
tirthasheshpatel: Changed this way, thanks @tqchen!

leofang (commented):
On my cell, sorry for brevity. Technically this note changes producers' required runtime behavior, should we bump the DLPack version as a heads-up?

tqchen (approved):

rgommers:
> Technically this note changes producers' required runtime behavior, should we bump the DLPack version as a heads-up?

I don't think it does really - implicitly this was always the requirement (because non-native endianness isn't handled at all), it just wasn't documented.

tirthasheshpatel:
Thanks for the reviews @edgarriba @tqchen @rgommers.

leofang:
Thanks, @tirthasheshpatel!

PR #101, opened 2022-03-25
[ABI break] Add new structs with version info and readonly flag

### LINKED ISSUE #34: Future ABI compatibility

### LINKED ISSUE #97: Endianness support is missing

### LINKED ISSUE #99: Clarify alignment requirement
DLPack has this comment in the header file: 

```
  /*!
   * \brief The data pointer points to the allocated data. This will be CUDA
   * device pointer or cl_mem handle in OpenCL. It may be opaque on some device
   * types. This pointer is always aligned to 256 bytes as in CUDA. The
   * `byte_offset` field should be used to point to the beginning of the data.
   *
   * Note that as of Nov 2021, multiply libraries (CuPy, PyTorch, TensorFlow,
   * TVM, perhaps others) do not adhere to this 256 byte aligment requirement
   * on CPU/CUDA/ROCm, and always use `byte_offset=0`.  This must be fixed
   * (after which this note will be updated); at the moment it is recommended
   * to not rely on the data pointer being correctly aligned.
   * ...
   */
```

This was discussed in https://github.com/data-apis/array-api/issues/293 and came up in NumPy https://github.com/numpy/numpy/issues/20338.

[This comment](https://github.com/data-apis/array-api/issues/293#issuecomment-964434449) by @rgommers summarizes the issue very well. Quoting the options in the comment:

> These are the options:
> 
> 1. **A1: required alignment**. Require the `data` pointer to always be aligned (using nonzero `byte_offset`), and do the gradual evolution plan in my comment above.
> 
> 2. **A2: no alignment**. remove the allocation requirement completely from `dlpack.h`. no library needs to make any changes (except if current handling of `byte_offset` is buggy, like @seberg pointed out for PyTorch). NumPy and other new implementers then just use `byte_offset=0` always (easiest), and we're done.
> 
> 3. **A3: optional alignment**. Do not _require_ alignment, but add a way to communicate from the producer to the consumer what the alignment of the data is.
> 
> 
> The current status is that the fine print in `dlpack.h` requires alignment (option A1), but no one adheres to it or enforces it. This state is not very useful: it requires a >1 year evolution plan, and apparently there's no gain because of the third bullet above. So it looks like the best choices are either A2 or A3. A3 seems strictly better than A2, _and most of the work it requires (versioning/extensibility) is work we wanted to do for other reasons already_.
> 
> So here's a new proposal:
> 
> * Decide that the long-term desired state is _A3: optional alignment_
> 
> * NumPy and other new implementers to do whatever is simplest, i.e. to use `byte_offset = 0` and `data` pointing to the first element in memory.
> 
> * Update the comment in `dlpack.h` about this topic to reflect: current state, desired future state, and a link to a new issue on the DLPack repo with more info (outcome of this discussion to be summarized on that issue).

I agree with @rgommers that **A3** is the best option because: most libraries don't care about alignment and we can communicate the alignment to some that do using the `__dlpack_info__` dunder or a request API under discussion in #34. If others agree, then let's add that to the spec and remove the comment from the header.

I noticed Tensorflow 2.8.0 crashed with NumPy 1.22.3 (on Ubuntu 20.04) due to an alignment issue:

```python
import numpy, tensorflow

x = numpy.arange(5, dtype='int64')
tf_x = tensorflow.experimental.dlpack.from_dlpack(x.__dlpack__())
tf_x[1]  # Fatal Failure
# 2022-03-15 18:29:10.587129: F ./tensorflow/core/framework/tensor.h:776] Check failed: IsAligned()
# Aborted
```

This happens for `float16`, `float32`, and `int64`.

(This crash doesn't happen for Tensorflow 2.7.0 and NumPy 1.22.3 but for all other versions, it does)

So, it'd be good to communicate alignment to the importing library. I don't think it'd be difficult to calculate that at importer side but it could be useful for opaque pointers.

tirthasheshpatel:
Closes #34 
Supersedes #72

This PR adds version info to the `DLTensor` struct. This is under discussion in #34 (see [this comment](https://github.com/dmlc/dlpack/issues/34#issuecomment-1078962112)). Since this is an ABI break, the ABI version has been incremented to 2 and the DLPack Version has been bumped to v0.7.0. This also updates the Python spec with a detailed upgrade policy and future ABI compatibility. The upgrade policy has been designed such that the libraries that want to keep supporting (exporting and importing) the old ABI or support multiple DLPack and ABI versions can do so without breaking compatibility with the old Python API.


include/dlpack/dlpack.h
```diff
@@ -200,6 +209,15 @@ typedef struct {
...
   uint64_t byte_offset;
+  /*! \brief Mark the data readonly. */
+  uint8_t readonly;
+  /*!
+  * \brief Endianness of the data. 1 for non-native endianness and
+  *  0 for native endianness.
+  */
+  uint8_t endianness;
+  /*! \brief Alignment of the data. */
+  uint32_t alignment;
```
tirthasheshpatel, line 220: These changes still need to be discussed. Since we are breaking the ABI by adding the version info, it would be good to discuss these here and add them too (so we don't need to break the ABI again)

include/dlpack/dlpack.h
```diff
@@ -158,6 +158,15 @@ typedef struct {
...
  */
 typedef struct {
+  /*!
+  * \brief The DLPack and ABI versions of the tensor. The exporting library
+  *  should set the DLPack version to DLPACK_VERSION and ABI version to
+  *  DLPACK_ABI_VERSION.
+  */
+  struct {
+    uint8_t dlpack;
+    uint8_t abi;
```
tqchen, line 168: Let us define DLPackVersion struct, for backward compatibility, consider put it in the end(after byte_offset), we need to consider the alignment properties of these fields.

tqchen: This will also make followup compiler support easier

docs/source/python_spec.rst
```diff
@@ -65,24 +66,27 @@ struct members, gray text enum values of supported devices and data
...
-capsule is no longer needed.
-
-The consumer must transer ownership of the ``DLManangedTensor`` from the
+``DLManagedTensor`` that is compatible with the DLPack and DLPack ABI
+version requested by the consumer. It will be consumed immediately
+within ``from_dlpack`` - therefore it is consumed exactly once, and it
+will not be visible to users of the Python API.
+
+The producer must set the ``PyCapsule`` name to ``"dltensor"`` if ABI
+version 1 is requested and ``"versioned_dltensor"`` if ABI version >= 2
```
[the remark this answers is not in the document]
tirthasheshpatel: > I asked in [#98 (review)](https://github.com/dmlc/dlpack/pull/98#pullrequestreview-903401339) and I will ask again: With all these changes flush in, shouldn't we bump the DLPack version at least one more time to include #98 #100 before breaking ABIs? Otherwise, ABI version is never set to 1 in any outside work; it only lives in the dev branch here, and this doc change doesn't really make much sense.

Oh, yes, thanks for noticing! I will propose a PR bumping the DLPack version and rebase here once that's merged.

leofang (changes requested):

wjakob:
An ABI break may also be an opportunity to reexamine the signed/unsignedness of all `DLTensor` members (PR #102).

include/dlpack/dlpack.h
tirthasheshpatel: Done, thanks!

docs/source/python_spec.rst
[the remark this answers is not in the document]
tirthasheshpatel: I have bumped the DLPack version here itself since it makes sense to upgrade both the DLPack and ABI version for every ABI break. Sounds good?

tirthasheshpatel:
Thanks for the help so far everyone! I think the discussion on #34 has settled on adding the `__dlpack_info__` method which returns the max version supported by the producer. I have updated the PR with the new spec (that includes `__dlpack_info__`). I haven't changed capsule renaming yet. @leofang Let us know if remaining the capsule makes sense to you when the new ABI is exported (as mentioned in #34, it provides some extra protection without breaking backward compatibility).

mattip:
I originally was in favor of capsule renaming, but I think it is not needed given the consensus around the protocol. While it does provide another layer of safety, I think it is redundant.

wjakob:
There are still a number of things I find confusing about the Python specification. Some of them are related to this specific PR. Some of them are things that are unclear in general.

1. The `stream` argument of `__dlpack__()` can presumably be used to pass a CUDA stream object. Is the interface specific to CUDA? If so, the documentation should probably say so.

CUDA streams are not a natural representation in CUDA. Even in C++, they are simply a pointer to an opaque NVIDIA data structure.  Are they passed as capsules, uintptr_t cast to a python int, etc.? The documented interface lacks type annotations to make this clear.

2. The document mentions a new `__dlpack_info__` function but does not provide a clear type-checkable signature of this new function. This is a recipe for inconsistencies. I recommend providing an example raw python version with MyPy-style type signature as an implementation guide.

3. What are the requirements on the `version` attribute of `__dlpack__`? Is it a Python `int`? If so, this could be stated in the type signature.

tirthasheshpatel:
> I originally was in favor of capsule renaming, but I think it is not needed given the consensus around the protocol. While it does provide another layer of safety, I think it is redundant.

Okay, I have removed capsule renaming now.

Also, I have added release notes and updated the DLPack diagram with the new structs. This PR is ready for review again from my side.

tqchen:
@tirthasheshpatel just realized that we checkedin the png image to the repo in a previous path. Is it possible to checkin the image to a different location/repo? On one hand we would like to keep repo self-contained, on the other hand, it can be a bit annoying for a repo to contain a rich history of binaries through multiple revisions. 

One approach is to send a PR to a separate repo instead (we used https://github.com/dmlc/web-data) for some of that purposes then link to via https://gitcontent

tqchen:
https://github.com/dmlc/dlpack/issues/104

docs/source/python_spec.rst
```diff
@@ -48,6 +50,14 @@ producer; the producer must synchronize or wait on the stream when necessary.
...
 unnecessary so asynchronous execution is enabled.
 
+A DLPack version can be requested by passing the ``version`` keyword. The
+consumer should call the ``__dlpack_info__`` method to get the maximum
+DLPack version supported by the producer and request for a version both
+support e.g. ``min(producer_version, consumer_version)``. If the consumer
+doesn't support any version below the producer's maximum version, a
+``BufferError`` should be raised. Similarly, If the producer doesn't
+support the requested version, it should raise a ``BufferError``.
+
```
mattip, line 60: What happens if the consumer does not specify a version? As I understand things, for the forseeable future the producer should return a V1 structure. So effectively the default is `1`.

wjakob:
> stream is an integer; the ID of the GPU device where the data is present. It is specific to CUDA and ROCm (and the devices that use a stream mechanism). This is explicitly stated in the [semantics section](https://dmlc.github.io/dlpack/latest/python_spec.html#semantics) of the python specification. Is it not clear enough?

I actually find this part of the documentation quite confusing (That said, I am only familiar with the CUDA and can't speak about ROCm).

There are three concepts in CUDA that play a role here:

- Each compatible graphics card/accelerator in a machine is given *CUDA device* ID.
- An application can talk to such a CUDA device by creating a *CUDA context* ([details](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__CTX.html#group__CUDA__CTX)). The CUDA runtime API creates a default context (_primary context_), but this is not necessarily only one. An application could in principle create multiple contexts to launch applications on the same GPU, each with its own virtual address space on the GPU (meaning that a memory allocation in one context cannot be accessed by another context).
- In each context, the application can create multiple *CUDA streams* ([details](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__STREAM.html#group__CUDA__STREAM)). A CUDA stream has the purpose of ordering kernel launches with respect to each other. It does *not* play a role with regards to the residency of memory allocations. (Although, to make this even more complicated, the latest CUDA versions now also have what is called stream-ordered memory allocations,I  digress...)

I think that there is I think a fundamental confusion between the word "stream" and "CUDA context" or "CUDA device ID" in the DLPack documentation. I suspect that will you want to call this "device ID" and not "stream" and also establish the convention that CUDA memory allocations are assumed to be located in the primary context associated with the associated CUDA device ID.

include/dlpack/dlpack.h
```diff
@@ -200,27 +210,60 @@ typedef struct {
...
   /*! \brief The offset in bytes to the beginning pointer to data */
   uint64_t byte_offset;
-} DLTensor;
+  /*! \brief The DLPack and DLPack ABI versions. The exporting library
+   *  should set the DLPack version to DLPACK_VERSION and ABI version to
+   *  DLPACK_ABI_VERSION. */
+  DLPackVersion version;
+  /*! \brief Mark the data readonly. */
+  uint8_t readonly;
+} DLTensorVersioned;
```
tqchen, line 219: Actually one thing that might be useful is to think about how to minimizing the change. Specifically, in light of S0 in https://github.com/dmlc/dlpack/issues/104. We may not need to introduce a separate versioned struct. 

The old framework can still use the DLTensor(as implemented in the versioned version) as it is, as long as there is no reliance of sizeof(DLTensor). Perhaps we can rename the old one as DLTensorLegacy; In the similar spirit, it is safe to `static_cast`  a DLTensor* to DLTensorLegacy*, as a result, we do not need to introduce two versions of DLManagedTensor

tqchen (changes requested):

docs/source/python_spec.rst
tirthasheshpatel: > What happens if the consumer does not specify a version? As I understand things, for the forseeable future the producer should return a V1 structure. So effectively the default is `1`.

Yes, the default is 1. I will update the spec to say that instead.

include/dlpack/dlpack.h
tirthasheshpatel: > The old framework can still use the DLTensor(as implemented in the versioned version) as it is, as long as there is no reliance of sizeof(DLTensor). Perhaps we can rename the old one as DLTensorLegacy; In the similar spirit, it is safe to `static_cast` a DLTensor* to DLTensorLegacy*, as a result, we do not need to introduce two versions of DLManagedTensor

But DLManagedTensor doesn't have a pointer to the DLTensor struct. Instead, it is just a plain `DLTensor dl_tensor` field. So, changing the sizeof(DLTensor) will change alignment and size of DLManagedTensor. Which is why we need to rename both structs. I also considered renaming the old structs to `DLTensorLegacy` and `DLManagedTensorLegacy` but it seems equivalent to instead add new structs with a different name (new implementations will be able to see both and can use the struct they want to export...)

tqchen: Ah sorry, you are indeed right. This would make the change ABI breaking. Although it would make future ABI change hard because all of them will be non-backward compact due to the change of deleter offset. Would be great to have more deliberation of this topic.

One possible choice is to append other flags(version, read_only) to DLManagedTensor instead. This of course won't be too ideal in case we want to change fields in DLTensor, but would maintain backward compact

docs/source/python_spec.rst
```diff
@@ -16,10 +16,12 @@ The array API will offer the following syntax for data interchange:
...
-2. ``__dlpack__(self, stream=None)`` and ``__dlpack_device__`` methods on the
-   array object, which will be called from within ``from_dlpack``, to query
-   what device the array is on (may be needed to pass in the correct
-   stream, e.g. in the case of multiple GPUs) and to access the data.
+2. ``__dlpack__(self, stream: int = None, version: int = None)``,
+   ``__dlpack_info__(self)``, and ``__dlpack_device__(self)`` methods
+   on the array object, which will be called from within ``from_dlpack``,
+   to access the data, to get the maximum supported DLPack version, and
+   to query what device the array is on (may be needed to pass in the
+   correct stream, e.g. in the case of multiple GPUs).
```
mattip, line 24: Can you break this into 3 separate points? They can be subpoints of 2

docs/source/python_spec.rst
```diff
@@ -96,7 +112,7 @@ C API which is called either when the refcount on the capsule named
       PyObject *type, *value, *traceback;
       PyErr_Fetch(&type, &value, &traceback);
 
-      DLManagedTensor *managed = (DLManagedTensor *)PyCapsule_GetPointer(self, "dltensor");
+      DLManagedTensorVersioned *managed = (DLManagedTensorVersioned *)PyCapsule_GetPointer(self, "dltensor");
```
mattip, line 115: How can we distinguish between these two cases? For legacy capsules, we must cast to the older struct, right?

docs/source/python_spec.rst
```diff
@@ -136,6 +153,44 @@ refer to `github.com/dmlc/dlpack <https://github.com/dmlc/dlpack>`_.
...
+    supporting it) or
+  * a ``BufferError`` should be raised (if the producer doesn't want to keep
+    support for the old ABI)
+
+  Otherwise, a capsule named ``"dltensor"`` which uses the new ABI
+  (``DLTensorVersioned`` and ``DLManagedTensorVersioned``) should be returned.
+  If the requested version is not supported, ``__dlpack__`` should raise a
+  ``BufferError``.
+* Producers should implement a ``__dlpack_info__`` method that returns the
+  maximum supported DLPack version.
```
mattip, line 185:
```suggestion
* Producers should implement a ``__dlpack_info__`` method that returns the
  maximum supported DLPack version. If this method does not exist, the consumer
  must use the old ABI.
```

docs/source/python_spec.rst
tirthasheshpatel: > How can we distinguish between these two cases? For legacy capsules, we must cast to the older struct, right?

The offset of `dl_tensor.device` doesn't change with this ABI breaking change (and we also don't anticipate such a change in the future). So, casting it to either struct and accessing the field should work.

include/dlpack/dlpack.h
mattip: Ahh, I see. If we move them to DLManagedTensor, we can ensure the location of the deleter function is still at the same offset. Then the code 
```
DLManagedTensorVersioned *managed = (DLManagedTensorVersioned *)PyCapsule_GetPointer(self, "dltensor");
if (managed->deleter) {
         managed->deleter(managed);
```

docs/source/python_spec.rst
tirthasheshpatel: I added this information in the point below this one (that point is all about what the consumer should do). Does that look good or do you think we should add the point here?

include/dlpack/dlpack.h
tirthasheshpatel: Although not ideal, I too think it'd be better to add the new fields to the DLManagedTensor instead. This would eliminate the requirement of the `*Versioned` structs and also make it easier to implement support for the new ABI.

docs/source/python_spec.rst
mattip: Ahh, this is code created by the __producer__, not the consumer. When the consumer calls `obj.__dlpack__`, the producer creates a capsule, and the capsule's deleter function consists of the code here.

tqchen:
closed in favor of #113  

We would like to thanks @tirthasheshpatel for your effort on bringing in this change onward

Issue #104, opened 2022-04-05
[DISCUSS][RFC] DLPack Versioning and ABI Update


tqchen:
This is a thread to bring awareness and discuss the recent set of proposed change of adding versioning information to DLPack. 

### Summary

Up until now the DLPack C struct itself does not carry ABI version information, and we always maintain backward compatibility. Up until now there is no ABI breaking changes.
 
The overall stability is one reason why frameworks adopt DLPack in the first place and we would like to continue to do moving forward. In the meantime, it is indeed helpful to clarify the DLPack ABI version in the data structure itself.  contains a draft proposal for the change. We can still 

In short, we are looking into the possibility of attaching version information to the DLTensor struct, to minimize the ABI change, the version information is appended in the end of the current DLTensor

### Summary of key changes

- C struct: append version struct in the end of current DLTensor, we can still discuss the specific impl of version struct. There is also a possibility of readonly flag/mask in light of the change
- Python API: introduce `__dlpack_info__` function that returns the current supported API/ABI version.

### Compatibility implications

One thing that is worth clarifying is that whether this constitute a ABI breaking change. It depends on how people use it. Let us clarify the following scenarios.

- S0: Frameworks that consume the old DLPack ABI should be able to continue consume the new DLTensor and DLManagedTensor by ignoring the fields
     -  Specifically, the memory layout of the existing fields remains the same, so old access code should work out of the box
     -  The deleter signature remains the same, which means call into the deleter of DLManagerTensor(where deleter is defined by producer that produces the new DLTensor) should safely delete the new structure.
- S1: Framework that produces old DLPack ABI Tensor can be consumed by the new framework, if the new framework queries the existence of `__dlpack_info__` then do the conversion according
    - Note the catch is that if a new framework was handled an arbitrary (old) DLTensor, querying the ABI field will result in an undefined behavior. So the guard on the python side exporter is needed.
- S2: Interestingly, the old framework can include thew new DLTensor/DLManagedTensor struct, not updating any of its exchange implementations. From consumer pov, they can receive and use DLTensor as if it is an old struct.
    - The deleter will ensure proper deletion. 
    - It is fine as a result for framework to "soft update" to include the new header, but not updating `__dlpack_info__` to indicate that it is already at a new version

Normally S0 means that the data structure ABI is still backward compatible (new data structure being used in old scenarios). S1 somewhat seats in the future-compatible regime(old data structure being used in new scenarios).

### Related threads
[...]

PR #105, opened 2022-04-15
[ABI break] Append new fields to DLManagedTensor

tirthasheshpatel:
Alternative to #101
Addresses #34 and #104
Implements the A1 alternative in https://github.com/dmlc/dlpack/issues/104#issuecomment-1090423607

Instead of appending the new fields to the `DLTesor` which changes the layout of `DLManagedTensor` in a backward incompatible manner, we can add the fields at the end of `DLManagedTensor` itself. This way we won't require new structs to maintain backward compatibility.

On Python front: should we add the `__dlpack_info__` dunder and a version keyword as proposed in #101? I'd say yes because, without knowing the requested version, the producer (exporter) cannot tell if it can export a certain type of array e.g. NumPy throws an error if one tries to export a readonly array with the old ABI. On the consumer (importer) side, it can't tell which fields are safe to access without knowing what version is being imported.

cc @tqchen @mattip @seberg @rgommers @leofang

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -16,10 +16,10 @@
 #endif
 
 /*! \brief The current version of dlpack */
-#define DLPACK_VERSION 60
+#define DLPACK_VERSION 70
 
 /*! \brief The current ABI version of dlpack */
-#define DLPACK_ABI_VERSION 1
+#define DLPACK_ABI_VERSION 2
 
 /*! \brief DLPACK_DLL prefix for windows */
 #ifdef _WIN32
@@ -222,6 +222,12 @@ typedef struct DLManagedTensor {
    *   The destructors deletes the argument self as well.
    */
   void (*deleter)(struct DLManagedTensor * self);
+  /* The version of the DLManagedTensor */
+  int32_t version;
+  /*! \brief Mark the data readonly. */
+  uint8_t readonly;
+  /*! \brief 128-bytes of padding to preserve ABI compatibility in the future. */
+  uint8_t __padding[128];
 } DLManagedTensor;
 #ifdef __cplusplus
 }  // DLPACK_EXTERN_C
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
 # Set variables:
 #   * PROJECT_NAME
 #   * PROJECT_VERSION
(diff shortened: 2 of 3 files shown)


include/dlpack/dlpack.h
```diff
@@ -222,6 +222,12 @@ typedef struct DLManagedTensor {
    *   The destructors deletes the argument self as well.
    */
   void (*deleter)(struct DLManagedTensor * self);
+  /* The version of the DLManagedTensor */
+  int32_t version;
```
tirthasheshpatel, line 226: Instead of introducing a new struct with a DLPack and ABI version, we can just export the DLPack version since it uniquely determines the ABI version. (cc @mattip)

include/dlpack/dlpack.h
```diff
@@ -222,6 +222,12 @@ typedef struct DLManagedTensor {
...
+  /*! \brief Mark the data readonly. */
+  uint8_t readonly;
+  /*! \brief 128-bytes of padding to preserve ABI compatibility in the future. */
+  uint8_t __padding[128];
```
tirthasheshpatel, line 230: Is 128 bytes of padding enough (or is it too much)?

mattip:
I think this need to include the `__dlpack_info__` parts of #101 and to provide two structs: a v1 `DLManagedTensor` and a v2+ `DLManagedTensorVersioned`. The default will be to return a `DLManagedTensor`. If the consumer requests a higher version we can return a DLManagedTensorVersioned`. Then it is the consumer's responsibility to check the extra fields iff they requested the higher version.

tirthasheshpatel:
> I think this need to include the `__dlpack_info__` parts of #101 and to provide two structs: a v1 `DLManagedTensor` and a v2+ `DLManagedTensorVersioned`. The default will be to return a `DLManagedTensor`. If the consumer requests a higher version we can return a `DLManagedTensorVersioned`. Then it is the consumer's responsibility to check the extra fields iff they requested the higher version.

I thought the reason to append new fields to `DLManagedTensor` itself was to avoid the requirement of two different structs. Since the layout of `DLManagedTensor` or `DLTensor` doesn't change, there should not be any breaks in existing code if the same struct is exported with extra fields. Although, I agree we should add the `__dlpack_info__` and version keyword (as we did in #101) so importers know what fields are safe to access and exporters know if they can export a certain type of array.

tqchen:
closed in favor of #113  

We would like to thanks @tirthasheshpatel for your effort on bringing in this change onward

PR #106, opened 2022-05-02
Add tests for numpy, tensorflow, torch, and mxnet

tirthasheshpatel:
This PR adds basic tests for some python packages that support DLPack. Currently, it only tests CPU arrays. Some observations

- Tensorflow doesn't provide a `from_dlpack` function and both Tensorflow and MXNet don't have the `__dlpack__`, `__dlpack_version__` dunder methods. These need to be updated.
- Tensorflow fails to import some NumPy arrays (most probably a bug in the deletion mechanism). These tests have been skipped for now.

More tests can be added in follow-up PRs. These tests are triggered for each commit on main/pr. Should we instead schedule them to run every week or month? GPU isn't available so I am not sure if we can test libraries like CuPy on the CI. Is is possible to emulate GPU or other devices in the CI?

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -33,3 +33,9 @@ jobs:
     - name: Test
       run: |
         ./tests/scripts/task_build.sh
+    - name: Install Python Packages
+      run: |
+        python3 -m pip install -U pytest numpy torch tensorflow mxnet
+    - name: Python Tests
+      run: |
+        pytest -v
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@
 *~
 build
 bin
+
+# Python
+__pycache__
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,39 @@
+import tensorflow as tf
+import mxnet as mx
+
+dtypes = [
+    'uint8', 'uint16', 'uint32', 'uint64',
+    'int8', 'int16', 'int32', 'int64',
+    'float16', 'float32', 'float64',
+    'complex64', 'complex128'
+]
+
+arrays = [
+    [1, 2, 3],    # 1D array
+    1,            # ndim = 0 array
+    [],           # empty array
+    [[1, 2, 3],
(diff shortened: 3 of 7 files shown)


mattip:
It would be nice to get some eyes on this. Once there are tests for the current behaviour using different libraries, it should be easier to judge the impact of the changes in #104

asmeurer:
I'm not sure how the DLPack test suite should work in relation to the array API tests test suite. My initial thought is that it makes sense for DLPack to have its own test suite, since DLPack is independent of the array API. I believe we currently don't have any DLPack tests in the array API test suite beyond just basic testing if the methods exist and run without erroring. If DLPack itself is tested extensively by its own test suite, then perhaps the array API test suite can just defer to that, or pull it in somehow. 

CC @honno @rgommers @kgryte

leofang:
I am thinking the tests from this PR (and future PRs) should have been added to each library's test suite when they claim they support DLPack. Without a universal test in this repo that can target every library I worry that we're just reinventing the wheel here... But, there's no way to write a universal test without an agreed-upon interface, and the array API could serve as a good proxy.

tirthasheshpatel:
To add onto @leofang's comment, for some given library (e.g. NumPy), it is not easy to test the DLPack support against other tensor libraries (e.g. PyTorch) since it would add extra dependencies for the project and affect the CI time.

leofang:
Sorry I didn't read Tirth's code until now. This PR essentially proposes to do an N-to-N conversion test across N libraries. The test suite would be hard to maintain unless 1. we use the array API to be backend-agnostic, **_or_** 2. we have an automation to auto-generate the added test scripts.

mattip:
Perhaps in the future, if N becomes large, it might be hard to maintain the tests. For now, with N = 4 there are ~16 tests. Each library has a different pattern for using DLPack, so it is not easy to see how to automate it today. It would be good to put this in now, and automate it when N grows larger and the use of DLPack becomes more standardized.

tirthasheshpatel:
@tqchen Any thoughts on this?

PR #114, opened 2022-11-10
Add `kDLBool` type

### LINKED ISSUE #75: [DISCUSS] Support kDLBool type
To represent tensors of type bool - is it possible adding to the `DLDataTypeCode` enum another type (kDLBool)? That can be very useful, as some well known machine learning platforms (TF, torch, Onnxruntime) support models whose inputs/outputs are tensors of type bool. Thanks!

leofang:
Close #75. Close #76. Supersedes #76.

It turns out we all forgot @alonre24 has already pushed a PR (#76), so all I did was minor edits in the docstrings and sync with the latest master, with his commit preserved.

cc: @tqchen


alonre24:
> Close #75. Close #76. Supersedes #76.
> 
> It turns out we all forgot @alonre24 has already pushed a PR (#76), so all I did was minor edits in the docstrings and sync with the latest master, with his commit preserved.
> 
> cc: @tqchen

LGTM
Glad that this suggestion has finally approved :)

include/dlpack/dlpack.h
```diff
@@ -134,10 +136,11 @@ typedef enum {
...
  *
  *  Examples
- *   - float: type_code = 2, bits = 32, lanes=1
- *   - float4(vectorized 4 float): type_code = 2, bits = 32, lanes=4
- *   - int8: type_code = 0, bits = 8, lanes=1
+ *   - float: type_code = 2, bits = 32, lanes = 1
+ *   - float4(vectorized 4 float): type_code = 2, bits = 32, lanes = 4
+ *   - int8: type_code = 0, bits = 8, lanes = 1
  *   - std::complex<float>: type_code = 5, bits = 64, lanes = 1
+ *   - bool: type_code = 6, bits = 8, lanes = 1
```
tqchen, line 143: would be good to add more comments: bool comes with bits=8 as it is normally the underlying storage type of common array libraries.

leofang: Done in [fd9b858](https://github.com/dmlc/dlpack/pull/114/commits/fd9b85822816caee1da480a300d841f0ff6f6e5b).

include/dlpack/dlpack.h
```diff
@@ -126,6 +126,11 @@ typedef enum {
    * (C/C++/Python layout: compact struct per complex number)
    */
   kDLComplex = 5U,
+  /*!
+   * \brief boolean
+   * The underlying storage size of bool is 8 bits, as per common array library convention
```
tqchen, line 131: it is actually better to annotate in another location(where i commented). Since here it can cause confusion that Bool always is 8bits regardless of the bit field. But in reality the current convention can support things like Bool(bits=1, lanes=32) for 32 bit bitmask

leofang: I would think the examples are meant to be canonical, so for example we would not allow alternative interpretations for common types like `float4` (and `bool`), otherwise it'd destroy the purpose of having an agreed upon format to facilitate exchange?

tqchen: That is fair, the main concern i had about the current one is that a reader may expect `Bool(bits=1, lane=1)` to have 8 bits. Instead, i would write we expect bool to be `Bool(bits=8, lanes=1)` as a canonical way of storage.

I am not stating that `Bool(bits=1, lanes=32)` should be supported right away, just avoid confusion and leave doors for future compatibility

leofang: I see, thanks for clarification. Updated.

include/dlpack/dlpack.h
leofang: Updated again.

tqchen (approved):

leofang:
@tqchen shall we merge and make a new release (before #113 is merged)?

seberg:
@tqchen was it correct that if we use the new Bool dtype with a size of 1-bit, than the storage would be assumed to actually be packed into a single bit?
Or was the intention that bit-size is padded to full bytes?!

There is some need for passing bit-masks around in the dataframe community, so it would be good to clarify this use-case, if valid.

tqchen:
in this case i think it should be ideally represented as `Bool(bit=1, lanes=32)`, which represent a 32bit bitmask.

I am not too sure if we want to specify  `Bool(bit=1, lanes=1)`, since that can be machine dependent. On a 8bit-byte machine  I guess we might still want to pad to minimum byte.  But say if a machine have bit-level addressing then it would be onebit.

seberg:
@tqchen to be honest, I have always had trouble understanding how lanes are to be used.  If I have say a 1111 elements in my array and bits does it work to say that `size=1111` but `dtype=Bool(bit=1, lanes=8)` (or something larger, since we mainly want to signal that its byte stored clearly)?

tqchen:
The lane represents the lanes of the unit-data type.

Say we want to store a bit mas, which is represented by int32. To store 65 bits, we will need 3 integers, in this case, it is 

`array(dtype=Bool(bits=1, lanes=32), size=3)` 

in the low level, so we have 3  * 32 bits in total.

If we store bool as a normal byte, and to store 65 bools, we need

`array(dtype=Bool(bits=8, lanes=1), size=65)`

seberg:
@tqchen but that is a problem, because how do I pass the 65 bits information there since `3 * 32 > 65` and the 65 is vital information!?

seberg:
I guess that the use-case I was asking for *could* use/abuse "lanes", because there is a side-channel to pass the *actual* shape.  But it doesn't feel ideal to me, so I am wondering if we can think of a pragmatic way to make this possible.
(To be honest, I have never seen "lanes" used, or dtypes relying on being padded to byte storage.)

seberg:
Maybe it would make sense to either introduce a new dtype or some sort of flag for bitmasks?

PR #133, opened 2023-12-11
Copy treatment on 0-size arrays from Python docs

leofang:
To make the C & Python docs in sync, as suggested in https://github.com/apache/arrow/pull/38472#discussion_r1417576828. cc: @pitrou @rgommers


rgommers (approved):
Thanks for doing this @leofang

pitrou (approved):

kkraus14 (approved):
Thanks @leofang

Issue #103, opened 2022-04-04
Interaction of `DLManagedTensor::deleter` and the GIL


wjakob:
Dear DLPack authors,

DLPack is widely used to enable interoperability between large C++-based frameworks that furthermore provide Python bindings. The C++ parts of such a framework are often usable through an independent C++ API (e.g. in the case of PyTorch, Tensorflow, ..), where functions can be called from multi-threaded code. In contrast, multithreading in Python is much more restricted: any use of the CPython API requires that the Global Interpreter Lock (GIL) is held. 

This discrepancy has implications on projects mixing C++ and Python code. Function in such a mixed C++/Python codebase should clarify require whether the caller should ensure that the GIL is held.

This is currently _unclear_ in the case of `DLManagedTensor::deleter`, which could be called by the `PyCapsule` destructor (where the GIL is held) or from arbitrary C++ code at some later point following "consumption" of the capsule (where the GIL is not necessarily held -- this destructor call could even occur from a different thread!)

I don't have any strong opinions either way, but ideally the documentation of the interfaces should say so.

Thanks,
Wenzel

tqchen:
Thanks @wjakob ! Indeed this is worth clarifying.

As far as I know, GIL is not necessary if the deleter is defined in C API. There can be cases where the deleter itself is defined by python's FFI mechanism, in such case, the GIL holding could happen inside the deleter, but nevertheless not the job of the caller of the deleter

rgommers:
The C docs clearly should not know anything about the GIL. As for the Python docs, this section talks about transfer of ownership of `DLManangedTensor` to the consumer: https://dmlc.github.io/dlpack/latest/python_spec.html#implementation. The `dlpack_capsule_deleter` directly uses the Python C API, so clearly already holds the GIL. The deleter can only be called once within `dlpack_capsule_deleter` (there's an early exit at the top). So it seems pretty clear at the moment. 

@wjakob what exact comment/phrasing would help you here?

wjakob:
How about the following?

> When the capsule has been consumed, the consumer library is responsible for calling ``DLManagedTensor::deleter``. From the viewpoint of the producer, no guarantees are given with respect to the execution context of this subsequent function call. For example, the call could take place on a different thread, and the Python GIL may or may not be held. It is the responsibility of the ``deleter`` implementation to handle such conditions using synchronization, if needed.

rgommers:
[...]

PR #140, opened 2024-03-22
DOC: Document that GIL must be grabbed (and Py_IsInitialized())

seberg:
I honestly don't remember if that ``Py_IsInitialized`` wasn't just to tape over C++ related deficiencies, but I suspect there isn't much to be avoided there...

(I.e. if we take care of the GIL for the other library, we also have to do this check for them which would normally be their job.)

Closes gh-103

### DIFF
--- a/docs/source/python_spec.rst
+++ b/docs/source/python_spec.rst
@@ -135,7 +135,8 @@ Note: the capsule names ``"dltensor"`` and ``"used_dltensor"`` must be
 statically allocated.
 
 The ``DLManagedTensor`` deleter must ensure that sharing beyond Python
-boundaries is possible, this means that the GIL must be acquired explicitly.
+boundaries is possible, this means that the GIL must be acquired explicitly
+if it uses Python objects or API.
 In Python, the deleter usually needs to ``Py_DECREF()`` the original owner
 and free the ``DLManagedTensor`` allocation.
 For example, NumPy uses the following code to ensure sharing with arbitrary
--- a/docs/source/python_spec.rst
+++ b/docs/source/python_spec.rst
@@ -134,6 +134,36 @@ C API which is called either when the refcount on the capsule named
 Note: the capsule names ``"dltensor"`` and ``"used_dltensor"`` must be
 statically allocated.
 
+The ``DLManagedTensor`` deleter must ensure that sharing beyond Python
+boundaries is possible, this means that the GIL must be held.
+In Python, the deleter usually needs to ``Py_DECREF()`` the original owner
+and free the ``DLManagedTensor`` allocation.
+For example, NumPy uses the following code to ensure sharing with arbitrary
+C++ is safe:
+
+.. code-block:: C
+   static void array_dlpack_deleter(DLManagedTensor *self)
+   {
+      /*
+       * Leak the pyobj if not initialized.  This can happen if we are running
+       * exit handlers that are destructing c++ objects with residual (owned)
+       * PyObjects stored in them after the Python runtime has already been
+       * terminated.
+       */
+      if (!Py_IsInitialized()) {
+         return;
+      }
+
+      PyGILState_STATE state = PyGILState_Ensure();
+
+      PyObject *array = (PyObject *)self->manager_ctx;
+      // This will also free the shape and strides as it's one allocation.
+      PyMem_Free(self);
+      Py_XDECREF(array);
+
+      PyGILState_Release(state);
+   }
+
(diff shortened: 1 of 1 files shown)


docs/source/python_spec.rst
```diff
@@ -134,6 +134,36 @@ C API which is called either when the refcount on the capsule named
 Note: the capsule names ``"dltensor"`` and ``"used_dltensor"`` must be
 statically allocated.
 
+The ``DLManagedTensor`` deleter must ensure that sharing beyond Python
+boundaries is possible, this means that the GIL must be held.
```
pitrou, line 138:
```suggestion
boundaries is possible, this means that the GIL must be acquired explicitly.
```

docs/source/python_spec.rst
```diff
@@ -134,6 +134,36 @@ C API which is called either when the refcount on the capsule named
...
+In Python, the deleter usually needs to ``Py_DECREF()`` the original owner
+and free the ``DLManagedTensor`` allocation.
+For example, NumPy uses the following code to ensure sharing with arbitrary
+C++ is safe:
```
pitrou, line 142:
```suggestion
non-Python code is safe:
```

docs/source/python_spec.rst
```diff
@@ -134,6 +134,36 @@ C API which is called either when the refcount on the capsule named
...
+
+.. code-block:: C
+   static void array_dlpack_deleter(DLManagedTensor *self)
+   {
+      /*
+       * Leak the pyobj if not initialized.  This can happen if we are running
+       * exit handlers that are destructing c++ objects with residual (owned)
+       * PyObjects stored in them after the Python runtime has already been
+       * terminated.
```
pitrou, line 151: This seems a bit over-specific.
```suggestion
       * Leak the Python object if the Python runtime is not available.
       * This can happen if the DLPack consumer destroys the tensor late
       * after Python runtime finalization (for example in case the tensor
       * was indirectly kept alive by a C++ static variable).
```

seberg: Thanks, much nicer, had just lazily copied it :).

pitrou:
> I honestly don't remember if that `Py_IsInitialized` wasn't just to tape over C++ related deficiencies

Our experience in PyArrow (not in DLPack context) is that it's required for robust integration with C++:
https://github.com/apache/arrow/blob/51817917e1436c8799ed382d160798060cd76652/python/pyarrow/src/arrow/python/common.h#L194-L199

docs/source/python_spec.rst
```diff
@@ -134,6 +134,36 @@ C API which is called either when the refcount on the capsule named
...
 
+The ``DLManagedTensor`` deleter must ensure that sharing beyond Python
+boundaries is possible, this means that the GIL must be acquired explicitly.
+In Python, the deleter usually needs to ``Py_DECREF()`` the original owner
+and free the ``DLManagedTensor`` allocation.
+For example, NumPy uses the following code to ensure sharing with arbitrary
+non-Python code is safe:
+
+.. code-block:: C
+   static void array_dlpack_deleter(DLManagedTensor *self)
```
pitrou, line 145: You might want to add a newline here to appease the reST parser?
```suggestion

   static void array_dlpack_deleter(DLManagedTensor *self)
```

pitrou: See CI error:
https://github.com/dmlc/dlpack/actions/runs/8387657417/job/22970268932?pr=140#step:5:156

seberg:
Yeah, makes sense:  If you knew it came from Python, it would be your job to store it in a way like that `OwnedRef` which deals with it.  But the whole issue here is that we want to allow you "forget" that this came from Python...

leofang (approved):

docs/source/python_spec.rst
```diff
@@ -134,6 +134,37 @@ C API which is called either when the refcount on the capsule named
 Note: the capsule names ``"dltensor"`` and ``"used_dltensor"`` must be
 statically allocated.
 
+The ``DLManagedTensor`` deleter must ensure that sharing beyond Python
+boundaries is possible, this means that the GIL must be acquired explicitly.
```
tqchen, line 138: aquired explicitly if the original memory comes from python runtime and deleter involves interacting with python runtime.

tqchen (changes requested):

docs/source/python_spec.rst
seberg: Fair, added "[...] if it uses Python objects or API"

tqchen (approved):

PR #142, opened 2024-03-23
Add hyperlinks to array api docs

### LINKED ISSUE #139: Circumnavigation in the Python DLPack docs
The page https://dmlc.github.io/dlpack/latest/python_spec.html currently lacks a detailed description of the DLPack Python protocol (for example the `__dlpack__` signature). It claims that "More details about the spec can be found under the [Data interchange mechanisms](https://data-apis.org/array-api/latest/design_topics/data_interchange.html#data-interchange) page", so one is tempted to follow through that link. However, the linked page does not provide a detailed description of the Python protocol either, and does not link to the actual [API docs](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html).

leofang:
Close #139.

Also fix the intra-crossrefs.


leofang:
cc: @pitrou for vis

PR #158, opened 2025-02-13
Add fp8/fp6/fp4 types

### LINKED ISSUE #156: [Feature Request] FP8 Support
It would be great to have FP8 support for converting tensors from PyTorch to DLPack. Currently, both PyTorch and TVM support FP8, but there is no direct way to convert tensors between them. Adding this support would improve interoperability and usability.

leofang:
Close #156. 

As discussed in the Python Array API meeting last week (Feb 6), I am adding additional data type enumerators based on user requests from, e.g., #156, https://github.com/pytorch/pytorch/issues/146414, and NVIDIA internal teams.

cc @tqchen @rgommers @seberg @oleksandr-pavlyk @hawkinsp @jakevdp @kmaehashi @kkraus14 @ptrblck @nouiz

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -157,6 +157,20 @@ typedef enum {
   kDLComplex = 5U,
   /*! \brief boolean */
   kDLBool = 6U,
+  /*! \brief FP8 data types. Must set DLDataType.bits=8. */
+  kDLFloat8_e3m4 = 7U,
+  kDLFloat8_e4m3 = 8U,
+  kDLFloat8_e4m3b11fnuz = 9U,
+  kDLFloat8_e4m3fn = 10U,
+  kDLFloat8_e4m3fnuz = 11U,
+  kDLFloat8_e5m2 = 12U,
+  kDLFloat8_e5m2fnuz = 13U,
+  kDLFloat8_e8m0fnu = 14U,
+  /*! \brief FP6 data types. Must set DLDataType.bits=6. */
+  kDLFloat6_e2m3fn = 15U,
+  kDLFloat6_e3m2fn = 16U,
+  /*! \brief FP4 data types. Must set DLDataType.bits=4. */
+  kDLFloat4_e2m1fn = 17U,
 } DLDataTypeCode;
 
 /*!
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -193,6 +193,9 @@ typedef enum {
  *   - float8_e4m3: type_code = 8, bits = 8, lanes = 1 (packed in memory)
  *   - float6_e3m2fn: type_code = 16, bits = 6, lanes = 1 (packed in memory)
  *   - float4_e2m1fn: type_code = 17, bits = 4, lanes = 1 (packed in memory)
+ *
+ *  When a sub-byte type is packed, DLPack requires the data to be in little bit-endian, i.e.,
+ *  for a packed data set D ((D >> (i * bits)) && bit_mask) stores the i-th element.
  */
 typedef struct {
   /*!
@@ -302,6 +305,14 @@ typedef struct DLManagedTensor {
  * consumer, until the producer-provided deleter is invoked.
  */
 #define DLPACK_FLAG_BITMASK_IS_COPIED (1UL << 1UL)
+
+/*
+ * \brief bit mask to indicate that whether a sub-byte type is packed or padded.
+ *
+ * The default for sub-byte types (ex: fp4/fp6) is assumed packed. This flag can
+ * be set by the producer to signal that a tensor of sub-byte type is padded.
+ */
+#define DLPACK_FLAG_BITMASK_IS_SUBBYTE_TYPE_PADDED (1UL << 2UL)
 
 /*!
  * \brief A versioned and managed C Tensor object, manage memory of DLTensor.
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -157,7 +157,7 @@ typedef enum {
   kDLComplex = 5U,
   /*! \brief boolean */
   kDLBool = 6U,
-  /*! \brief FP8 data types. Must set DLDataType.bits=8. */
+  /*! \brief FP8 data types */
   kDLFloat8_e3m4 = 7U,
   kDLFloat8_e4m3 = 8U,
   kDLFloat8_e4m3b11fnuz = 9U,
@@ -166,10 +166,16 @@ typedef enum {
   kDLFloat8_e5m2 = 12U,
   kDLFloat8_e5m2fnuz = 13U,
   kDLFloat8_e8m0fnu = 14U,
-  /*! \brief FP6 data types. Must set DLDataType.bits=6. */
+  /*! \brief FP6 data types
+   * Setting bits != 6 is currently unspecified, and the producer must ensure it is set
+   * while the consumer must stop importing if the value is unexpected.
+   */
   kDLFloat6_e2m3fn = 15U,
   kDLFloat6_e3m2fn = 16U,
-  /*! \brief FP4 data types. Must set DLDataType.bits=4. */
+  /*! \brief FP4 data types
+   * Setting bits != 4 is currently unspecified, and the producer must ensure it is set
+   * while the consumer must stop importing if the value is unexpected.
+   */
   kDLFloat4_e2m1fn = 17U,
 } DLDataTypeCode;
 
@@ -184,6 +190,9 @@ typedef enum {
  *   - int8: type_code = 0, bits = 8, lanes = 1
  *   - std::complex<float>: type_code = 5, bits = 64, lanes = 1
  *   - bool: type_code = 6, bits = 8, lanes = 1 (as per common array library convention, the underlying storage size of bool is 8 bits)
+ *   - float8_e4m3: type_code = 8, bits = 8, lanes = 1 (packed in memory)
+ *   - float6_e3m2fn: type_code = 16, bits = 6, lanes = 1 (packed in memory)
+ *   - float4_e2m1fn: type_code = 17, bits = 4, lanes = 1 (packed in memory)
  */
 typedef struct {
   /*!


include/dlpack/dlpack.h
```diff
@@ -157,6 +157,20 @@ typedef enum {
...
+  kDLFloat8_e4m3 = 8U,
+  kDLFloat8_e4m3b11fnuz = 9U,
+  kDLFloat8_e4m3fn = 10U,
+  kDLFloat8_e4m3fnuz = 11U,
+  kDLFloat8_e5m2 = 12U,
+  kDLFloat8_e5m2fnuz = 13U,
+  kDLFloat8_e8m0fnu = 14U,
+  /*! \brief FP6 data types. Must set DLDataType.bits=6. */
+  kDLFloat6_e2m3fn = 15U,
+  kDLFloat6_e3m2fn = 16U,
```
tqchen, line 171: likely we need to specify what happens when bits * lanes do not perfectly divide word side (32).

include/dlpack/dlpack.h
```diff
@@ -157,6 +157,20 @@ typedef enum {
   kDLComplex = 5U,
   /*! \brief boolean */
   kDLBool = 6U,
+  /*! \brief FP8 data types. Must set DLDataType.bits=8. */
+  kDLFloat8_e3m4 = 7U,
+  kDLFloat8_e4m3 = 8U,
+  kDLFloat8_e4m3b11fnuz = 9U,
```
tqchen, line 163: useful to elaborate a bit on description of each type, and comon usage, e.g. e8m0 can be used in scenarios like microscaling factor in gemm

include/dlpack/dlpack.h
```diff
@@ -157,6 +157,20 @@ typedef enum {
...
+  kDLFloat8_e4m3fn = 10U,
+  kDLFloat8_e4m3fnuz = 11U,
+  kDLFloat8_e5m2 = 12U,
+  kDLFloat8_e5m2fnuz = 13U,
+  kDLFloat8_e8m0fnu = 14U,
+  /*! \brief FP6 data types. Must set DLDataType.bits=6. */
+  kDLFloat6_e2m3fn = 15U,
+  kDLFloat6_e3m2fn = 16U,
+  /*! \brief FP4 data types. Must set DLDataType.bits=4. */
+  kDLFloat4_e2m1fn = 17U,
```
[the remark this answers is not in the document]
hawkinsp: I'll note that `numpy` has no support for sub-byte types and it seems like it might be a significant engineering project to add it. We certainly couldn't figure out how to do that for `ml_dtypes` and resorted to just keeping things padded up to 8 bits per element. Perhaps it is useful to support both packed and unpacked formats?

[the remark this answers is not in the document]
tqchen: would be good to know about the current use-case of fp6, perhaps it makes sense for us to phase it and leave it out for now. Since we know f4 is easier(as multiple of it aligns to bytes) and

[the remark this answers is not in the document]
tqchen: interesting. My guess the most common use-case would be packed variant for the F4. One possible way to state it could be make F4 and lanes=2, which makes it byte level, unfortunately that would leave lanes=1 unspecified.

include/dlpack/dlpack.h
tqchen: would also be good to discuss what is a good use-case of F6, since unlike F4, multiple of it does not fully map into a word (32bits),

seberg: Can't we just define (if nobody needs/uses them, nobody will support them).  And beyond clarifying that packing is defined by the `bits` field everything else needs to be figured out by the consumer (if we want to pass alignment/padding info, I don't think the dtype is the place to put it).

isVoid: FP6 doesn't bring a whole lot of performance benefits. It's largely memory saving on weights due to bitwidth reduction by 25%.

tqchen: the main thing we want to be careful here is create a spec then later come back wanting a different one. So that is why i think it could be good to phase it a bit for controversial ones

tqchen: @isVoid 's suggestion seems to indicate that we would primarily want packed layout for such case

tqchen:
Summarizing the current discussion. The main sticky pt as of now is how do we deal with sub-byte data types, and/or sub-byte datatype whose multiple is not aligned. Two situations:

- For F6, it is less for performance gain more more designed for memory saving, it is sub-byte, and its multiple do not aligns to a word.
- For F4, its multiple aligns to a byte, and there is performance gains leveraged by accelerated instructions from latest GPU.

Then main question is when we have subbyte data type while lane=1, what is the underlying data storage. There are two options

- A0: Padded, each element should be padded to byte-level
- A1: Packed, each element should be packed continuously in-memory without padding.

For performance and memory saving reasons, likely A1 is more desirable. However, frameworks like numpy may not be designed to support sub-byte level, mainly likely due to strides specification is in unit of  bytes instead of in unit of element(which DLPack adopts). So for certain implementations the array may indeed be A0.

During discussions @seberg also suggested that we could potentially use the `bits` field to indicate padding
- A2: When `bits=8` and we are looking at F4 for example, it indicated padding to bytes. but as @seberg suggested it is a bit patch on the protocol 

Alternatively, we can have a byte padding flag in the DLManagedTensor to indicate byte-level padding.
-A3: Having a byte_padding flag in DLManagedTensor to indicate that we would like to pad, and defaults to `False`(mainly because most use-cases are A1 for acceleration)

The main rationale that I think we should take are two folds
- Enable the latest high performance accelerated use-cases in array API
- Design minimal and future compatible specs(e.g. if we specify in one and and end up the future end up wanting another way it would be back)

Given the importance of F8 and F4, I think we should expedite enabling them. And most motivating cases are for A1.  I feel we could first go with A1, then A3(or A2, which is compatible with A1). We can also phase the stages so F6 comes after F4, given there might be a bit more discussion there(but they can also land very quickly once we converged on padding spec). 

Let me know if that would be a reasonable approach

include/dlpack/dlpack.h
leofang: This is not something I am knowledgeable enough to answer. I raised an internal discussion with the teams that requested the addition of fp6.  One of us can copy back the understanding here.

also note that ml_dtypes already supports fp6, so @hawkinsp might have some thoughts here 🙂

seberg: That sounds fine/good to defer!  Please change the `Must set DLDataType.bits=4` language to something like:
Setting DLDataType.bits != 4 is currently unspecified so the producer must ensure it is set and the consumer must stop importing if the value is unexpected.

leofang: Hi all, reporting back based on what I learned internally (but with public references). 

As far as DLPack is concerned (exchanging data residing in global memory), for fp6 we want 16 (sixteen) fp6 values packed into a 12B buffer, so that we can apply the `cp.async.bulk.tensor` PTX instruction to load the data from global memory to shared memory (and vice versa), see [this figure](https://docs.nvidia.com/cuda/parallel-thread-execution/#tensor-dimension-size-format-sub-bytes-padding-align-b6-16-p32) and [this figure](https://docs.nvidia.com/cuda/parallel-thread-execution/#tensor-dimension-size-format-sub-bytes-padding-align-b6-p2-16) from the PTX ISA documentation.

I think it is clear we want `bits=6` (@seberg I am confused what your `bits=4` comment was about?) but I am not sure if we should also require `lanes=16`?

seberg: Sorry, I slipped in the lines and was just looking at the `fp4`s below.

I would find prescribing `lanes=16` odd, because e.g. `(4, 4)` `fp6` array can be copied as a single 16 lane `fp6`, but can't be described by a 16 lane data-type.

In practice, maybe that is what anyone needs for fp6 (work with multiples of 16), so it seems perfectly good to use it.
Right now, everyone in Python just checks `lanes == 1`... but it would be easy enough to just do: `shape[i] = shape[i] * lanes` and `strides[i] = strides[i] / lanes`...  And if that is only really used in `fp6` world, so be it.

In principle a different mechanism to indicate alignment and padding would seem nicer to me, but we just don't have it right now.  So that `lanes == 1`, but the importer can easily figure out that using 16 lane instructions is perfectly fine.

leofang: Sounds good, I'll apply `lanes=1` to the comment.

include/dlpack/dlpack.h
```diff
@@ -157,6 +157,20 @@ typedef enum {
...
   kDLBool = 6U,
+  /*! \brief FP8 data types. Must set DLDataType.bits=8. */
+  kDLFloat8_e3m4 = 7U,
+  kDLFloat8_e4m3 = 8U,
+  kDLFloat8_e4m3b11fnuz = 9U,
+  kDLFloat8_e4m3fn = 10U,
+  kDLFloat8_e4m3fnuz = 11U,
+  kDLFloat8_e5m2 = 12U,
+  kDLFloat8_e5m2fnuz = 13U,
+  kDLFloat8_e8m0fnu = 14U,
```
szha, line 168: Should we reserve some places between the different float widths in case there are more variants that come up? I didn't get the sense that everyone converged on the format yet given so many variants.

tqchen: giveb enum is not ordered, i think we can go with the common ones and add more in future numbers

leofang: I thought about that too but unfortunately we only have 256 slots due to the use of `uint8_t` storage in `DLDataType`, so I tried to be cautious about leaving holes.

seberg:
So, I am OK with either, but it seems like not padding is the more useful thing for now.

Re-using `.bits` here does would lead to two spellings for a packed `int6` (not a biggy but true).
I think right now, we also never said what an `int` with `bits=2` is (i.e. is it padded to byte size or not?).

One other thing I am not sure about common use:  For sub-byte dtypes you also need to store a `bit_offset` (or re-interpret the `byte_offset` as bits).
If I think of NumPy supporting bit-sized, slicing + view semantics mean that you need a bit-offset, but likely nobody needs it for these (or even for bit-fields e.g. in dataframe world).

include/dlpack/dlpack.h
leofang: I don't think it is DLPack's responsibility to educate developers which type to pick.

leofang:
Thanks all. In commit 46122b1 I made some updates to ensure we unblock the A1 use cases (packed layouts), while leaving room for padded layouts in the future.

@hawkinsp @seberg my question would be: Even if DLPack allows both packed and padded layouts in the future, would NumPy's dtype registration mechanism allow ml_dtypes to register this info for NumPy to pick up during exchange through DLPack? IIUC NumPy's current DLPack implementation does not query these information from the dtype object (likewise, it does not query from the memory allocator the memory type). It seems like another gap to me (in addition to NumPy not supporting subtypes) that we need to consider sooner than later?

leofang:
FWIW @tqchen and I discussed offline. If everyone is OK with this PR, let's merge it and cut a v1.1 release.

tqchen (approved):

tqchen:
approving, let us wait for another three days to give everyone a chance to chime in

seberg:
> Even if DLPack allows both packed and padded layouts in the future, would NumPy's dtype registration mechanism allow ml_dtypes to register this info for NumPy to pick up during exchange through DLPack?

The problem is representing arbitrary sub-byte offsets and strides in the array object mainly.  Having padded vs. non-padded is in principle fully supported in NumPy.

tqchen:
@seberg @hawkinsp @szha @isVoid please see if you can take another look

seberg:
I am OK with this, my only concern is always possible future regret.  There might a regret in avoiding having two new flags rather than just one (i.e. a bit-sized indicator and a padding indicator).
But, since padding might make sense more broadly, I half suspect we need that either way if it becomes a serious use-case.

leofang:
How about defining `DLPACK_FLAG_BITMASK_IS_SUBBYTE_TYPE_PADDED` (default it to false) and requiring consumers to check the flag when accessing fp6/fp4/int4/...?

szha (approved):
LGTM

seberg:
> How about defining `DLPACK_FLAG_BITMASK_IS_SUBBYTE_TYPE_PADDED` (default it to false)

Right, but I feel that if we add a flag, it should likely also indicate that there is a new `bitoffset` field (or byte-offset has a bit-offset meaning), so that one can represent a `arr_f4[1:5]` which has a `ptr + 4bits` initial point.

Similarly for a "padded" flag:  Not sure that "padded to next byte" is the only thing one would need.

---

So unless we are sure about indicating e.g. `byteoffset -> bitoffset` flag, it seems just as well to put this in as is?

include/dlpack/dlpack.h
```diff
@@ -157,6 +157,26 @@ typedef enum {
...
+   * Setting bits != 6 is currently unspecified, and the producer must ensure it is set
+   * while the consumer must stop importing if the value is unexpected.
+   */
+  kDLFloat6_e2m3fn = 15U,
+  kDLFloat6_e3m2fn = 16U,
+  /*! \brief FP4 data types
+   * Setting bits != 4 is currently unspecified, and the producer must ensure it is set
+   * while the consumer must stop importing if the value is unexpected.
+   */
+  kDLFloat4_e2m1fn = 17U,
```
hawkinsp, line 179: Is it unambiguous which order the two nibbles appear within a byte? Does everyone agree on that?

tqchen: this is a good point, @leofang , we should explicitly state it. My understanding is that we can go with something conisstent with "little endian", 

```
for a byte B,  

((B >> (i * 4)) && bit_mask) stores the i-th element
```

tqchen:
@hawkinsp brought up a good point about pack ordering, which is related to endianess. So explicitly state it here to bring awareness


```
for a byte B,  

((B >> (i * 4)) && bit_mask) stores the i-th element
```

Note that such order is consist with little endian in general, when you have a big packed data set D and shift/masking operation defined for D.

```
((D >> (i * 4)) && bit_mask) stores the i-th element
```

is consistent for D equals one byte and D equals multiple bytes

tqchen:
I think it is reasonable to go with `DLPACK_FLAG_BITMASK_IS_SUBBYTE_TYPE_PADDED` and default to false. We can either also do it in a followup

include/dlpack/dlpack.h
leofang: Done in commit [a0a1597](https://github.com/dmlc/dlpack/pull/158/commits/a0a15970176116ba8f9c916a5d428fd0fbe499ce), following the suggestion below (https://github.com/dmlc/dlpack/pull/158#issuecomment-2685111237).

leofang:
> I think it is reasonable to go with `DLPACK_FLAG_BITMASK_IS_SUBBYTE_TYPE_PADDED` and default to false. We can either also do it in a followup

Done in commit a0a1597.

tqchen:
thanks @leofang ,i think all outstanding comments are addressed, going to merge before monday if nobody objetcts

tqchen:
Thanks everyone, this change is now merged

PR #165, opened 2025-05-12
Update minimum cmake required

### LINKED ISSUE #162: CMake 4.0 support
Please add CMake 4.0 support. We need to update the first line of the CMakeLists.txt file from:
```
cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
```
to 
```
cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
```
because CMake 4.0 has dropped the support for cmake version <3.5.

vyasr:
This PR updates the minimum required version of CMake. This change is necessary because the CMake 4.0 release [removed policies that allowed supporting such old versions of CMake](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features), so having such a low required minimum version of CMake prohibits using the latest versions. CMake 3.16 is already quite old (the release was [back in 2019](https://www.kitware.com/cmake-3-16-0-available-for-download/)) and it is a reasonable choice based on Linux distro support (for example, [it is the supported version for Ubuntu's 20.04 LTS release](https://packages.ubuntu.com/search?suite=focal&searchon=names&keywords=cmake)).

Resolves #162

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,10 @@
 ###
-# Set minimum version of CMake. Since command 'project' use
-# VERSION sub-option we need at least 3.0.
-# Note: If you use 2.6 or 2.4, God kills a kitten. Seriously.
-cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+# Set minimum version of CMake. CMake 4.0 drops support for CMake<3.06 so that
+# is the absolute lowest that we could allow without blocking the usage of new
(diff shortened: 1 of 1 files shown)


leofang (approved):

leofang:
Thanks, Vyas! The CI failure is known for some time and irrelevant of this PR. I dunno why #162 was not closed by #163 (which bumped the minimum to 3.5), though. But I agree even 3.5 is very old and we can raise the bar a bit further.

PR #171, opened 2025-09-04
Use commit hash to download lint script for CI failure

take-cheeze:
Since it was deleted in https://github.com/dmlc/dmlc-core/pull/693


leofang (approved):
Thanks, @take-cheeze! As a stop-gap solution this is totally fine.

@tqchen we need to switch to use another linter in the long term.

PR #174, opened 2025-09-12
[RFC] Support DLPACK C Functions for Speed Exchange and Stream Handling

tqchen:
This PR adds support for three C functions to speedup DLPack exchange. As of now, DLPack exchange relies on python functions such as `tensor.__dlpack__()`.

While they works well for common cases, the general overhead of such exchange is at the level of 0.2-0.3 us for very well optimized version, and can go up to 0.4-1 us for less optimized implementation.

For a function that takes three arguments f(a, b, c), assume we run DLPack exchange for each argument, the general conversion overhead usually gets to around 1us and sometimes to 3us.

While such overhead can be acceptable in many settings, in GPU applications the extra 1-3us overhead can still be significant.

This PR proposes four functions for speed exchange DLPack tensors without going through python interpreter.

- DLPackManagedTensorFromPyObjectNoSync for  fast exchange owned tensors into consumer
- DLPackDLTensorFromPyObjectNoSync for fast exchange non-owned tensors into consumer
- DLPackManagedTensorToPyObjectNoSync for fast exchange consumer tensors into producer (return values)
- DLPackCurrentWorkStream for query the current working stream from the stream context

Our preliminary results show that these functions, when incorporated correctly via native extensions such as c/c++, can bring exchange cost to the level of 30ns - 80ns, giving us about one order of maginitude speedup. That means the API overhead of functions like f(a, b, c) will be at 0.2us-0.4us level (including exchange), which is close to what native cpp extension overhead do without exchange.

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -1,5 +1,5 @@
 /*!
- *  Copyright (c) 2017 by Contributors
+ *  Copyright (c) 2017 -  by Contributors
  * \file dlpack.h
  * \brief The common header of DLPack.
  */
@@ -326,7 +326,7 @@ typedef struct DLManagedTensor {
  *
  * \note This is the current standard DLPack exchange data structure.
  */
-struct DLManagedTensorVersioned {
+typedef struct DLManagedTensorVersioned {
   /*!
    * \brief The API and ABI version of the current managed Tensor
    */
@@ -360,7 +360,86 @@ struct DLManagedTensorVersioned {
   uint64_t flags;
   /*! \brief DLTensor which is being memory managed */
   DLTensor dl_tensor;
-};
+} DLManagedTensorVersioned;
+
+//--------------------------------------------------------------------
+// DLPack C functions for speed exchange
+//--------------------------------------------------------------------
+/*
+ * \brief A generic C-style allocator that exposes allocation of a Tensor/Array.
+ *
+ * Array/Tensor libraries can store this field as an int in the type of the Tensor/Array.
+ *
+ * mypackage.Tensor.__c_dlpack_tensor_allocator__ = MyPackageDLPackTensorAllocator
+ *
+ * This information can then be used to set allocators of a callee to run allocations.
+ *
+ * This particular function does not assume a Python environment; as a result,
+ * the error handling mechanism is different from Python-related functions.
+ *
+ * \param prototype The prototype DLTensor to offer details about the device and shape.
+ *                  Other field information will be ignored during allocation.
+ * \param out The output DLManagedTensorVersioned.
+ * \param error_ctx The context to set the error.
+ * \param SetError The function to set the error.
+ * \return 0 on success, -1 on failure.
+ *         The callee should call SetError(error_ctx, kind, message) to set the error kind and message.
+ * \note Error propagation via SetError.
+ */
+typedef int (*DLPackTensorAllocator)(                                       //
+  DLTensor* prototype, DLManagedTensorVersioned** out, void* error_ctx,     //
+  void (*SetError)(void* error_ctx, const char* kind, const char* message)  //
+);
+
+/*!
+ * \brief Exports a PyObject* Tensor/NDArray to a DLManagedTensorVersioned.
+ *
+ * This function is a C-style function pointer to quickly convert a PyObject* Tensor/NDArray
+ * to a DLManagedTensorVersioned without going through the Python Interpreter.
+ *
+ * It also provides an option to query the current context stream of the device provided
+ * by the tensor.
+ *
+ * Array/Tensor libraries can store this field as an int in the type of the Tensor/Array.
+ *
+ * mypackage.Tensor.__c_dlpack_from_pyobject__ = MyPackageDLPackFromPyObject
+ *
+ * This information can then be picked up by importers and libraries to run the speed conversion.
+ * This function should not throw any exceptions; if it fails, it should return -1 and
+ * set the error message via PyErr_SetXXX.
+ *
+ * \param py_object The Python object to convert; this should be PyObject*.
+ *                  We use void* to avoid dependency on Python.h.
+ * \param out The output DLManagedTensorVersioned.
+ * \param optional_out_env_stream Outputs the current context stream of the device provided
+ *                   by the tensor; it can be NULL, in which case the stream will not be queried.
+ * \return 0 on success, -1 on failure. PyError should be set if -1 is returned.
+ * \note We use void* to avoid dependency on Python.h, so this specific type is
+ *       not dependent on Python.h and can be copied to dlpack.h.
+ */
+typedef int (*DLPackFromPyObject)(                              //
+  void* py_object,                                              //
+  DLManagedTensorVersioned** out,                               //
+  void** optional_out_env_stream                                //
+);
+
+/*!
+ * \brief Imports a DLManagedTensorVersioned to a PyObject* Tensor/NDArray.
+ *
+ * This function is a C-style function pointer to quickly convert a DLManagedTensorVersioned
+ * to a PyObject* without going through the Python Interpreter.
+ *
+ * Array/Tensor libraries can store this field as an int in the type of the Tensor/Array.
+ *
+ * mypackage.Tensor.__c_dlpack_to_pyobject__ = MyPackageDLPackToPyObject
+ *
+ * \param tensor The DLManagedTensorVersioned to convert.
+ * \param out_py_object The output Python object.
+ * \return 0 on success, -1 on failure. PyError should be set if -1 is returned.
+ * \note We use void* to avoid dependency on Python.h, so this specific type is
+ *       not dependent on Python.h and can be copied to dlpack.h.
+ */
+typedef int (*DLPackToPyObject)(DLManagedTensorVersioned* tensor, void** out_py_object);
 
 #ifdef __cplusplus
 }  // DLPACK_EXTERN_C
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -407,7 +407,16 @@ typedef int (*DLPackManagedTensorAllocator)(                                //
  *
  * \param py_object The Python object to convert; this should be PyObject*.
  *                  We use void* to avoid dependency on Python.h.
+ *
+ * \param max_version The maximum version of DLPack support that consumer supports.
+ *                    Consumer should fill in their own version here, this parameter is not null.
+ *                    Producer can use this information to produce the appropriate
+ *                    DLManagedTensorVersioned for maximum compatibility if needed.
+ *                    This field is primarily used for future compatibility in case
+ *                    of major version bump and ABI-breaking changes.
+ *
  * \param out The output DLManagedTensorVersioned.
+ *
  * \param optional_out_env_stream Outputs the current context stream of the device provided
  *                   by the tensor; it can be NULL, in which case the stream will not be queried.
  *                   optional_out_env_stream should points to cudaStream_t in the case of CUDA.
@@ -420,6 +429,7 @@ typedef int (*DLPackManagedTensorAllocator)(                                //
  */
 typedef int (*DLPackManagedTensorFromPyObject)(                 //
   void* py_object,                                              //
+  const DLPackVersion* max_version,                             //
   DLManagedTensorVersioned** out,                               //
   void** optional_out_env_stream                                //
 );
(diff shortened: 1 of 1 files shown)


tqchen:
RFC https://github.com/dmlc/dlpack/issues/175

tqchen:
updates to incorporate suggestion by @dalcinl.

include/dlpack/dlpack.h
```diff
@@ -360,6 +360,137 @@ struct DLManagedTensorVersioned {
...
+ *     version.minor = DLPACK_MINOR_VERSION;
+ *     tensor_allocator = MyDLPackTensorAllocator;
+ *     dlpack_from_py_object = MyDLPackFromPyObject;
+ *     dlpack_to_py_object = MyDLPackToPyObject
+ *  }
+ *
+ *  const DLPackExchangeAPI* Global() {
+ *     static MyDLPackExchangeAPI inst;
+ *     return &inst;
+ *  }
```
oleksandr-pavlyk: I believe this method needs to be static. This is to make `MyDLPackExchangeAPI::Global()` usage valid.

tqchen: good catch

tqchen:
This PR is updated to reflect the all the suggestions in https://github.com/dmlc/dlpack/issues/175

Naming (updated per suggestion from @oleksandr-pavlyk @kkraus14):
- DLPackManagedTensorAllocator
- DLPackManagedTensorToPyObject 
- DLPackManagedTensorFromPyObject 

Clarified that the dunder should be attached to the class type @gbonik and @seberg

include/dlpack/dlpack.h
```diff
@@ -360,6 +360,150 @@ struct DLManagedTensorVersioned {
...
+ * \return 0 on success, -1 on failure. PyError should be set if -1 is returned.
+ * \note We use void* to avoid dependency on Python.h, so this specific type is
+ *       not dependent on Python.h and can be copied to dlpack.h.
+ *
+ * \sa DLPackExchangeAPI
+ */
+typedef int (*DLPackManagedTensorFromPyObject)(                 //
+  void* py_object,                                              //
+  DLManagedTensorVersioned** out,                               //
+  void** optional_out_env_stream                                //
```
kkraus14: I would really like to get fully aligned on #176 before we commit to passing streams like this

kkraus14 (commented):
Everything LGTM except for the ongoing discussions related to stream / synchronization handling

tqchen:
Thanks everyone for comments. We have updated the proposal to include a non-owned version while explicitly state clearly the intend is for fast caling for framework library/DSL call use cases

Summary of the current functions

- DLPackManagedTensorFromPyObjectNoSync for fast exchange owned tensors into consumer
- DLPackDLTensorFromPyObjectNoSync for fast exchange non-owned tensors into consumer
- DLPackManagedTensorToPyObjectNoSync for fast exchange consumer tensors into producer (return values)
- DLPackCurrentWorkStream for query the current working stream from the stream context

I think it is getting to a mergable state, but would be good to also get final inputs if any

yongwww:
I’m in favor of this RF. Adding C APIs for DLPack exchange and stream querying is a very compelling direction, achieving exchange latency down to tens of nanoseconds could remove a significant bottleneck. This RFC is a valuable addition to DLPack. I’m excited to see it get landed and deployed in production!

yongwww (approved):

tqchen:
incoprorated suggestions from @seberg into the PR. i think we are in good shape, going to merge this in two days if there is no more future comments.

Thanks everyone for suggestions so far

include/dlpack/dlpack.h
```diff
@@ -360,6 +360,253 @@ struct DLManagedTensorVersioned {
...
+ * \note - As a C function, must not thrown C++ exceptions.
+ *       - Error propagation via SetError to avoid any direct need
+ *         of Python API. Due to this `SetError` may have to ensure the GIL is
+ *         held since it will presumably set a Python error.
+ *
+ * \sa DLPackExchangeAPI
+ */
+typedef int (*DLPackManagedTensorAllocator)(                                         //
+  DLTensor* prototype, DLManagedTensorVersioned** out, void* error_ctx,              //
+  void (*SetError)(void* error_ctx, const char* kind, const char* message)           //
```
seberg, line 393: One small thing is, I am not sure how we will use `kind`?  Let's say this typically get's converted to a Python exception, then we would have to _somewhat_ agree on e.g. `MemoryError` or so to translate that.

Maybe this will just settle itself, in the end I bet this pretty much only ever raises `MemoryError` anyway.

(I think this function is important! I know I am slow to settle on something, but I haven't quite settled on loving this approach. But I suspect its true that asking to allocate e.g. the `torch` tensor here -- requiring the GIL -- and then viewing that is also not really better.)

include/dlpack/dlpack.h
```diff
@@ -360,6 +360,253 @@ struct DLManagedTensorVersioned {
...
+   * It must be NULL if the framework does not support older versions.
+   * If the current major version is larger than the one supported by the
+   * consumer, the consumer may walk this to find an earlier supported version.
+   *
+   * \sa DLPackExchangeAPI
+   */
+  struct DLPackExchangeAPI* prev_version_api;
+  /*!
+   * \brief Producer function pointer for DLPackManagedTensorAllocator
+   *        This function must be not NULL.
```
seberg:
```suggestion
   *        This function must not be NULL.
```
a bit clearer maybe? but silly nit. (also below)

include/dlpack/dlpack.h
```diff
@@ -360,6 +360,253 @@ struct DLManagedTensorVersioned {
...
+  DLPackManagedTensorFromPyObjectNoSync managed_tensor_from_py_object_no_sync;
+  /*!
+   * \brief Producer function pointer for DLPackManagedTensorToPyObject
+   *        This function must be not NULL.
+   * \sa DLPackManagedTensorToPyObject
+   */
+  DLPackManagedTensorToPyObjectNoSync managed_tensor_to_py_object_no_sync;
+  /*!
+   * \brief Producer function pointer for DLPackDLTensorFromPyObject
+   *        This function can be NULL when the producer does not support this function.
```
seberg, line 612: I like this approach as such, but as I just mentioned it elsewhere, I want to point out that if this is optional and commonly not support, it means all consumers will need:
```
if (api->managed_tensor_to_py_object_no_sync == NULL) {
    // do complicated thing
}
```
which unfortunately means that we can get a small speed improvement if available (not having to allocate the `DLManagedTensor`, etc.) but we may need to support both paths as a consumer, unfortunately.

I suppose part of the solution here may be that you really want a C++ convenience layer that is easy to vendor... 


This isn't a deal breaker! It made me wonder if filling in a `DLManagedTensor` is much worse -- it would be different from currently by not owning its own allocation.

include/dlpack/dlpack.h
```diff
@@ -360,6 +360,253 @@ struct DLManagedTensorVersioned {
...
+ *
+ * Convert an owning DLManagedTensorVersioned* to the Python tensor of the
+ * producer (implementor) library with the correct type.
+ *
+ * This function does not perform any stream synchronization.
+ *
+ * This function is exposed by the framework through the DLPackExchangeAPI.
+ *
+ * \param tensor The DLManagedTensorVersioned to convert the ownership of the
+ *        the data is stolen.
```
seberg:
```suggestion
 *        tensor is stolen.
```
probably me, but more specific.

include/dlpack/dlpack.h
```diff
@@ -360,6 +360,253 @@ struct DLManagedTensorVersioned {
...
+ * \brief Obtain the current work stream of a device.
+ *
+ * Obtain the current work stream of a device from the producer framework.
+ * For example, it should map to torch.cuda.current_stream in PyTorch.
+ *
+ * When device_type is kDLCPU, the consumer do not have to query the stream
+ * and the producer can simply return NULL when queried.
+ * The consumer do not have to do anything on stream sync or setting.
+ * So CPU only framework can just provide a dummy implementation that
+ * always set out_current_stream[0] to NULL.
```
seberg, line 457: I forgot about the CPU part, allowing the function pointer to be `NULL` makes sense.  But I am also happy to just implement always `NULL` return or an always `-1` with a "NumPy doens't have streams, what's going on" error.

include/dlpack/dlpack.h
```diff
@@ -360,6 +360,253 @@ struct DLManagedTensorVersioned {
...
+ *
+ * \param device_type The device type.
+ * \param device_id The device id.
+ * \param out_current_stream The output current work stream.
```
seberg, line 461: I am a bit unsure if this is specified as well as it needs to be?  I.e. I think NULL would be the default stream.

The question is: Is there any need (or not) an "undefined or no synchronization" return value (such as -1)?
If not, we are all good, but if some producer might need this (for whatever reason), then we need to specify this here.

The alternative is that the producer just has to return the default stream (otherwise the consumer has to guess a stream anyway probably in the kernel use-case!).

seberg (approved):
I like this approach in general.  I am still a bit unsure about the allocation function and the filling one, but if nobody else chimes in, then so be it...

One question about the stream sync, we should clarify that a bit maybe.

One curiosity: Should there be an entry (or function?) that allows discovering which device types are supported?

---

I think in a sense, I still need to see how this actually looks like for a real-world consumer/producer pair, but in the near future it is maybe still malleable enough.

include/dlpack/dlpack.h
tqchen: i don;t think there is a need in this particular context. returning the default stream is likely more well defined

include/dlpack/dlpack.h
tqchen: yes, likely `MemoryError` is the most common one likely

tqchen:
thanks @seberg on entry function to discover device supported, as of now no, but if there is a need, we could add such API. although i am not sure if it is strictly needed

include/dlpack/dlpack.h
```diff
@@ -360,6 +360,253 @@ struct DLManagedTensorVersioned {
...
+  /*!
+   * \brief Optional pointer to an older DLPackExchangeAPI in the chain.
+   *
+   * It must be NULL if the framework does not support older versions.
+   * If the current major version is larger than the one supported by the
+   * consumer, the consumer may walk this to find an earlier supported version.
+   *
+   * \sa DLPackExchangeAPI
+   */
+  struct DLPackExchangeAPI* prev_version_api;
```
dalcinl: In the future, a version v2 may have a `DLPackExchangeAPI` struct with more/different entries than v1, therefore the struct corresponding to v2 may not have the same layout (and thus type name) than the older v1.

I assume that the intention here is that  `DLPackExchangeAPI` will grow, if ever, to the end; additional, the entries currently defined below will never change (not their relative position, nor the layout of each struct). 
I that correct? If not, this will require some nasty and inconvenient fixes.

Long story short: this is a great idea, but the implementation as it is now is not flexible to future changes the layout of the various structures. Just double checking we are all in the same page here.

tqchen: This is a good point. I think we can say only the two first two fields remains unchanged, will do a patch to clarify this

tqchen: updated to move the exchange api stable part to a new struct DLPackExchangeAPIHeader

dalcinl: If things ever change after the first two fields, then the C code users will write to deal with older versions will violate the C language strict aliasing rules and entering the undefined behavior territory.

dalcinl: Oh, I see you changed things to use a header.

tqchen:
thanks everyone for valuable feedbacks so far, planning to merge in 24hours

include/dlpack/dlpack.h
tqchen: @dalcinl if it works for you, would be great if you can explicitly approve, thanks

dalcinl (approved):

include/dlpack/dlpack.h
```diff
@@ -360,7 +360,266 @@ struct DLManagedTensorVersioned {
...
+  /*!
+   * \brief Optional pointer to an older DLPackExchangeAPI in the chain.
+   *
+   * It must be NULL if the framework does not support older versions.
+   * If the current major version is larger than the one supported by the
+   * consumer, the consumer may walk this to find an earlier supported version.
+   *
+   * \sa DLPackExchangeAPI
+   */
+  struct DLPackExchangeAPIHeader* prev_version_api;
```
dalcinl: Just a minor nit... This name containing both "version" and "api" may be slightly confusing,

What about just?
```suggestion
  struct DLPackExchangeAPIHeader* prev_api;
```
PS: Just a mild suggestion, I'm fine either way.

tqchen:
Thanks everyone, this is merged

Issue #177, opened 2025-09-30
[RFC] Enforce strides to be not NULL for non-zero ndims


tqchen:
In the past, we allow DLTensor.strides to be NULL and use that to signal compactness. In practice, most frameworks will always export with strides for nonzero ndims. Enforcing strides to be not NULL would help with memory safety here. So this RFC proposes for us to enforce the strides to be not NULL.

hpkfft:
If `ndim == 0`, then requiring both `shape` and `strides` to be `NULL` (instead of, for example, pointing to one past the end of some chunk of allocated memory) might be a good idea.

Regardless, as a general comment on the DLPack standardization process, I would very much like to see buy-in from the most important frameworks (e.g., NumPy) before anything is added to DLPack.
I'm not familiar with the NumPy code base, so I may be mistaken, but a cursory glance suggests to me that NumPy specifically sets `strides` to `NULL` if the tensor is compact and row-majored: [dlpack.c](https://github.com/numpy/numpy/blob/17d5deda759e3a3ce79d88966ce69fdb323f623b/numpy/_core/src/multiarray/dlpack.c#L290)

tqchen:
Agree on the case of `ndim==0` would be OK.


@rgommers @seberg  maybe can help comment on numpy pov

seberg:
No big opinion from me, Numpy always fills it in I think, but could also special case it for C contiguity. The current rule would mainly help frameworks that only have C contiguity, but without requiring NULL (so there is no need to confirm) its maybe just as well to force setting it.

rgommers:
The proposal seems quite reasonable to me, and I agree with @seberg that it should be fine for NumPy. Even if there's a case where it'd be NULL currently, it shouldn't be hard to change that.

seberg:
I'll just put it in to NumPy, since I doubt that anyone will complain.

Consumers can't rely on it anyway (it would be a minor speedup if they check for contiguous otherwise).

So the main annoyance will be for a producer that only has C-contiguous tensors.  That isn't NumPy, though.
You are basically just shifting the annoyance from such a producer to the consumer, but maybe there are fewer such producers.

tqchen:
seems like that we are converging towards adding it to the spec. I will send a PR about this

tqchen:
https://github.com/dmlc/dlpack/pull/178

leofang:
[...]

PR #178, opened 2025-10-09
Enforce strides to be not null when ndim is nonzero

tqchen:
This PR updates the spec to enforce the strides to be not null when the ndim is nonzero.
This is going to help consumers to handle cases more uniformly.

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -257,8 +257,9 @@ typedef struct {
   /*! \brief The shape of the tensor */
   int64_t* shape;
   /*!
-   * \brief strides of the tensor (in number of elements, not bytes)
-   *  can be NULL, indicating tensor is compact and row-majored.
+   * \brief strides of the tensor (in number of elements, not bytes),
+   *  can not be NULL if ndim != 0, must points to
+   *  an array of ndim elements that specifies the strides.
    */
   int64_t* strides;
   /*! \brief The offset in bytes to the beginning pointer to data */
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -254,12 +254,23 @@ typedef struct {
   int32_t ndim;
   /*! \brief The data type of the pointer*/
   DLDataType dtype;
-  /*! \brief The shape of the tensor */
+  /*!
+   * \brief The shape of the tensor
+   *
+   *  When ndim == 0, we suggest to set the shape to NULL.
+   */
   int64_t* shape;
   /*!
    * \brief strides of the tensor (in number of elements, not bytes),
    *  can not be NULL if ndim != 0, must points to
    *  an array of ndim elements that specifies the strides.
+   *
+   *  When ndim == 0, we suggest to set the strides to NULL.
+   *
+   *  \note Before DLPack v1.2, strides can be NULL to indicate contiguous data.
+   *        This is not allowed in DLPack v1.2 and later. The rationale
+   *        is to simplify the consumer hanlding, and most frameworks already
+   *        always returns the strides when ndim != 0.
    */
   int64_t* strides;
   /*! \brief The offset in bytes to the beginning pointer to data */
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -269,8 +269,7 @@ typedef struct {
    *
    *  \note Before DLPack v1.2, strides can be NULL to indicate contiguous data.
    *        This is not allowed in DLPack v1.2 and later. The rationale
-   *        is to simplify the consumer hanlding, and most frameworks already
-   *        always returns the strides when ndim != 0.
+   *        is to simplify the consumer handling.
    */
   int64_t* strides;
   /*! \brief The offset in bytes to the beginning pointer to data */
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -19,7 +19,7 @@
 #define DLPACK_MAJOR_VERSION 1
 
 /*! \brief The current minor version of dlpack */
-#define DLPACK_MINOR_VERSION 1
+#define DLPACK_MINOR_VERSION 2
 
 /*! \brief DLPACK_DLL prefix for windows */
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


include/dlpack/dlpack.h
```diff
@@ -257,8 +257,9 @@ typedef struct {
   /*! \brief The shape of the tensor */
   int64_t* shape;
   /*!
-   * \brief strides of the tensor (in number of elements, not bytes)
-   *  can be NULL, indicating tensor is compact and row-majored.
+   * \brief strides of the tensor (in number of elements, not bytes),
+   *  can not be NULL if ndim != 0, must points to
+   *  an array of ndim elements that specifies the strides.
```
hpkfft: I suggest **requiring** `NULL` if `ndim == 0`.   That is:
```
 *  which is NULL if and only if ndim == 0, pointing to
 *  an array of ndim elements that specify the strides.
```

junrushao (approved):

include/dlpack/dlpack.h
```diff
@@ -257,8 +257,9 @@ typedef struct {
   /*! \brief The shape of the tensor */
```
hpkfft: I suggest **requiring** NULL if `ndim == 0`.  That is:
```
 *  \brief The shape of the tensor, which is NULL if and only if ndim == 0,
 *   pointing to an array of ndim elements.
```

hpkfft (dismissed):

include/dlpack/dlpack.h
seberg: While I do think suggesting it makes sense, I am not sure I see a benefit (how can you rely on it)?

On a general note, we really can't just change this without leaving a comment?  From a purity point of view: before version 1.2 the strides were allowed to be `NULL` when the array was C contiguous.
I'll settle for something fuzzier, but I feel we need a paper trail (for at least a year or so)?

It's not like we know nobody does this!

tqchen: let us go with suggesting it instead of requiring it for now. Agree on @seberg to add the paper trail here

tqchen:
Updated to add the suggestion to set pointers to NULL when ndim==0, add notes about change in DLPack v1.2

tqchen:
https://github.com/dmlc/dlpack/issues/177

include/dlpack/dlpack.h
hpkfft: What do you mean by "it"?
I do not think it makes sense to *suggest* strides be NULL if ndim == 0 if you're going to *require* strides not to be NULL when ndim > 0.
Either suggest both (and then there is no need to increment the version number), or require both (in v1.2).

include/dlpack/dlpack.h
```diff
@@ -254,11 +254,23 @@ typedef struct {
...
+   * \brief strides of the tensor (in number of elements, not bytes),
+   *  can not be NULL if ndim != 0, must points to
+   *  an array of ndim elements that specifies the strides.
+   *
+   *  When ndim == 0, we suggest to set the strides to NULL.
+   *
+   *  \note Before DLPack v1.2, strides can be NULL to indicate contiguous data.
+   *        This is not allowed in DLPack v1.2 and later. The rationale
+   *        is to simplify the consumer hanlding, and most frameworks already
+   *        always returns the strides when ndim != 0.
```
hpkfft: Minor: "handling" is misspelled.
Do you have evidence that most frameworks already always return non-NULL strides when ndim != 0?
NumPy, for example, returns NULL for C-contiguous arrays before today's not-yet-released development version.
Even if the statement is true for most frameworks, it may not be useful to mention it in the header since consumers ought to correctly handle NULL anyway (e.g., to support the current and past releases of NumPy).

include/dlpack/dlpack.h
tqchen: I meant the ndim == 0 indicate strides == NULL. Like @seberg said, I agree it makes sense, however, also not sure see a benefit where one can rely on it(unlike the strides not null case), so would be good to know what are potential usecases here

seberg: I just don't see the reason to prescribe it. Forcing null seems like a code quality thing to to make sure things crash if incorrectly used and things aren't "undefined". Or do you have one?

include/dlpack/dlpack.h
tqchen: noted, updated the typo. And rmeoved the note on frameworks here

include/dlpack/dlpack.h
hpkfft: #177 seemed (originally, at least) to be motivated by memory safety (and, I would add, debugability).
In the case `ndim == 0`, either `strides` is NULL, or it is some garbage value (e.g., it points to unrelated allocated memory or it points to unallocated memory).   Well, I guess it could point to a new array of one element, but that seems unnatural and wasteful.   My question is, in the case `ndim == 0`, if `strides` is not NULL what other value would be reasonable/expected?

tqchen: Some implementation might do the following thing:

- Allocate DLTensor, plus the strides/shape inplace that follows the DLTensor
- `dltensor.shape =  reinterpret_cast<int64_t*>(reinterpret_cast<char*>(self) + sizeof(DLTensor))`
- `dltensor.strides =  dltensor.shape + dltensor.ndim`

This is actually quite convenient implementation and it is nice that it does not have to check for ndim = 0 and specially set shape/strides to NULL.

And because ndim == 0, it is safe in a sense that there is no memory value bein accessed from shape/strides

seberg: I would frame it as: the consumer can rely on `strides[dim]` working. Apparently that was desired.
To me that isn't much about memory safety, but maybe you can convince me that there is a serious memory safety angle somewhere additionally.

hpkfft: So, yeah, fundamentally just a code quality thing.
But I do think the clause "which is NULL if and only if ndim == 0" is clearer to humans than "cannot be NULL if ndim != 0".
From a verification point of view, it's easier to write a test suite when a standard specifies everything rather than when it leaves the value of `strides` unspecified in corner cases.
Just my two cents--I'm not trying to make trouble if others prefer to leave it unspecified.

hpkfft: If `ndim == 1`, then (in the future) the consumer can rely on `strides[0]` working.  Yes, I see that.

@seberg, in NumPy https://github.com/numpy/numpy/pull/29872, we now have:
```
  dl_tensor->shape = (ndim > 0) ? (int64_t *)((char *)ptr + offset) : NULL;
  dl_tensor->strides = (ndim > 0) ? dl_tensor->shape + ndim : NULL;
```
That seemed to me to be worth the minor inconvenience.
Otherwise, `shape` and `strides` would point past the end of allocated memory.
Is that still OK with you?

@tqchen, in the opening comment of #177, you wrote, "Enforcing strides to be not NULL would help with memory safety here. "   Would you be willing to further explain your motivation, and perhaps, what you feel is best from a memory safety point of view if `ndim ==0`?

tqchen: i somewhat agree with @seberg that consumers can rely on `strides[dim]` working for dim that `0<= dim < ndim`

tqchen: specificially, i think it is OK for `shape` and `strides`(and maybe desirable from the implementation simplicity pov) to point to the end of allocated memory for ndim=0 case

tqchen:
Thanks everyone for the feedbacks. We updated to explicit state the constraint 
`consumer can always rely on strides[dim] being valid for 0 <= dim < ndim.` 

This ensures that the standard still allows shape/strides to points to the last of the allocated memory when ndim=0, and not having to special case for the scenario

tqchen:
most outstanding comments are resolved, see also note above on the rationale on the ndim=0 case, i think we can plan to merge soon soon and have a 1.2 release. Thanks @hpkfft @seberg and everyone for chiming in

Kathryn-cat (approved):

tqchen:
planning to merge in 24hours

Issue #179, opened 2025-12-03
[RFC] DLPack C Exchange API Attribute to PyCapsule


tqchen:
Background in https://github.com/dmlc/dlpack/issues/175 we introduced the C exchange API. In the original discussion we have considered proposals of PyCapsule as field and in dunder.  We discussed the choice array-api meeting session where and I think everyone agreed both path can work, but then eventually went with int.

After recent discussion with PyTorch https://github.com/pytorch/pytorch/issues/162845#issuecomment-3609127149 there is a good feedback that having PyCapsule with name "dlpack_exchange_api" would provide better checking guarantees.

Given the original discussion everyone seems to agree both can work, I think it is good to take the additional feedback and update to bring PyCapsule support here. This RFC proposes to update the convention to allow the `__c_dlpack_exchange_api__` field to take PyCapsule (with name "dlpack_exchange_api").

tqchen:
cc @seberg @rgommers @leofang @dalcinl @albanD @ngimel

dalcinl:
I'm a bit confused. What exactly `__c_dlpack_exchange_api__` is supposed to be? I would argue is has to be a callable, such that ` capsule = exporter.__c_dlpack_exchange_api__()`, like the current `__dlpack[_device]__` dunders.

seberg:
It was discussed as a class attribute and not a method, since there should never be a reason to pass any arguments.
That said now that you mention it, a _class_ attribute is slightly more annoying to implement as it has to be put into the `tp_dict` in the C-API (overhead wise that might still be better since at least you can do a `getattr(x, "attr", NotImplemented)` to avoid an explicit `hasattr` check or errors (I can't think of a "call if exists" style API).

I don't see much of a downside to using a capsule, it seems largely a matter of taste. It does add a bit of safety, although it also adds a tiny bit of overhead if you use a capsule name, which is nice for safety. (Of course all of these overheads are like 10-20ns at most.)

tqchen:
It was discussed as a class attribute. This being said, in the PyTorch discussion the callable choice was indeed bought up by @junrushao . Explicitly listing out choices:

- A0a. Take `__c_dlpack_exchange_api__`  as a PyCapsule in class attribute
    - A0b: original    `__c_dlpack_exchange_api__` as class attribute and int (I think we should move away from this one, given it is in latest and not a lot of implementation, it is likely OK)
- A1. Update the  `__dlpack_c_exchange_api__` to be a method which returns a PyCapsule. 
[...]

PR #180, opened 2025-12-05
Update exchange API to be capsule following new convention

tqchen:
This PR updates the exchange API to be capsule following the new convention. Note that the exchange API python recommendation is not yet updated on the array-api side and so far there hasn't yet been downstream implementation, so this change should be safe in generally.

### DIFF
--- a/docs/source/python_spec.rst
+++ b/docs/source/python_spec.rst
@@ -208,8 +208,7 @@ and then upgrade to the C Exchange API later when faster data exchange is needed
    because the DLPack C exchange API operates at the C extension level, we need
    direct interaction between the array framework ``PyObject*`` and DLPack,
    as a result it is harder to implement the C Exchange API through ctypes (because
-   ctypes releases GIL by default and sometimes in non-free-threading environment,
-   GIL is needed to interact with the Python C API).
+   ctypes release thread state by default which is needed to interact with the Python C API).
 
 A reference implementations of the C Exchange API in frameworks:
 
--- a/docs/source/python_spec.rst
+++ b/docs/source/python_spec.rst
@@ -203,7 +203,7 @@ and then upgrade to the C Exchange API later when faster data exchange is needed
    Producer framework should implement the C Exchange API in a static way either
    through Cython, Python C extensions, or Python binding mechanism. Importantly,
    because the DLPack C exchange API operates at the C extension level, we need
-   direct interaction between the array framework PyObject* and DLPack,
+   direct interaction between the array framework ``PyObject*`` and DLPack,
    as a result it is harder to implement the C Exchange API through ctypes (because
    ctypes releases GIL by default and sometimes in non-free-threading environment,
    GIL is needed to interact with the Python C API).
--- a/docs/source/python_spec.rst
+++ b/docs/source/python_spec.rst
@@ -184,7 +184,7 @@ DLPack C Exchange API
 
 Starting with DLPack 1.3, a new C Exchange API is introduced to enable faster
 data exchange than the Python ``__dlpack__`` API at the C extension level.
-Producer array frameworks must provide a ``__dlpack_c_exchange_api__``
+Producer array frameworks may provide a ``__dlpack_c_exchange_api__``
 attribute on the array type.
 The attribute should be a ``PyCapsule`` with name ``"dlpack_exchange_api"``.
 The consumer can query whether this attribute exists and use it at the C extension level.
--- a/docs/source/python_spec.rst
+++ b/docs/source/python_spec.rst
@@ -193,9 +193,12 @@ and then upgrade to the C Exchange API later when faster data exchange is needed
 
 .. code-block:: C
 
-   PyObject *api_obj = type(tensor_obj).__dlpack_c_exchange_api__;  // as C code.
-   MyDLPackExchangeAPI *api = PyCapsule_GetPointer(api_obj, "dlpack_exchange_api");
-   if (api == NULL && PyErr_Occurred()) { goto handle_error; }
+   // Get type, fetch capsule attribute, and extract the C struct pointer
+   PyObject *api_capsule = PyObject_GetAttrString((PyObject *)Py_TYPE(tensor_obj), "__dlpack_c_exchange_api__");
+   if (api_capsule == NULL) { goto handle_error; }
+   MyDLPackExchangeAPI *api = (MyDLPackExchangeAPI *)PyCapsule_GetPointer(api_capsule, "dlpack_exchange_api");
+   Py_DECREF(api_capsule);
+   if (api == NULL) { goto handle_error; }
 
 
 .. note:: Implementation of the C Exchange API
--- a/docs/source/python_spec.rst
+++ b/docs/source/python_spec.rst
@@ -179,6 +179,42 @@ refer to `github.com/dmlc/dlpack <https://github.com/dmlc/dlpack>`_.
    guaranteed to be in a certain order or not.
 
 
+DLPack C Exchange API
+~~~~~~~~~~~~~~~~~~~~~
+
+Starting with DLPack 1.3, a new C Exchange API is introduced to enable faster
+data exchange than the Python ``__dlpack__`` API at the C extension level.
+Producer array frameworks must provide a ``__dlpack_c_exchange_api__``
+attribute on the array type.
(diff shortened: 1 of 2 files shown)


tqchen:
cc @Kathryn-cat

tqchen:
RFC thread https://github.com/dmlc/dlpack/issues/179

junrushao (approved):

Kathryn-cat (approved):

tqchen:
cc @leofang @rgommers also updated python_spec.rst here

tqchen:
Array api spec PR https://github.com/data-apis/array-api/pull/984

leofang:
Sorry for delay, will try to finish my review in a day or two!

docs/source/python_spec.rst
```diff
@@ -179,6 +179,42 @@ refer to `github.com/dmlc/dlpack <https://github.com/dmlc/dlpack>`_.
    guaranteed to be in a certain order or not.
 
 
+DLPack C Exchange API
+~~~~~~~~~~~~~~~~~~~~~
+
+Starting with DLPack 1.3, a new C Exchange API is introduced to enable faster
+data exchange than the Python ``__dlpack__`` API at the C extension level.
+Producer array frameworks must provide a ``__dlpack_c_exchange_api__``
```
leofang: I think in array API we follow RFC 2119 and it'd be better to reword this as
```suggestion
Producer array frameworks may provide a ``__dlpack_c_exchange_api__``
```

docs/source/python_spec.rst
```diff
@@ -179,6 +179,42 @@ refer to `github.com/dmlc/dlpack <https://github.com/dmlc/dlpack>`_.
...
+attribute on the array type.
+The attribute should be a ``PyCapsule`` with name ``"dlpack_exchange_api"``.
+The consumer can query whether this attribute exists and use it at the C extension level.
+Notably, consumer frameworks can always start implementing by only using the Python ``__dlpack__`` API,
+and then upgrade to the C Exchange API later when faster data exchange is needed.
+
+.. code-block:: C
+
+   PyObject *api_obj = type(tensor_obj).__dlpack_c_exchange_api__;  // as C code.
```
leofang: I am confused. Is this valid C code? Shouldn't we use `PyObject_HasAttrString` or alike to query if the _Python_ attribute exists with this type?

docs/source/python_spec.rst
```diff
@@ -179,6 +179,42 @@ refer to `github.com/dmlc/dlpack <https://github.com/dmlc/dlpack>`_.
...
+   MyDLPackExchangeAPI *api = PyCapsule_GetPointer(api_obj, "dlpack_exchange_api");
+   if (api == NULL && PyErr_Occurred()) { goto handle_error; }
+
+
+.. note:: Implementation of the C Exchange API
+
+   Producer framework should implement the C Exchange API in a static way either
+   through Cython, Python C extensions, or Python binding mechanism. Importantly,
+   because the DLPack C exchange API operates at the C extension level, we need
+   direct interaction between the array framework PyObject* and DLPack,
```
leofang: nit:
```suggestion
   direct interaction between the array framework ``PyObject*`` and DLPack,
```

docs/source/python_spec.rst
```diff
@@ -179,6 +179,42 @@ refer to `github.com/dmlc/dlpack <https://github.com/dmlc/dlpack>`_.
...
+   as a result it is harder to implement the C Exchange API through ctypes (because
+   ctypes releases GIL by default and sometimes in non-free-threading environment,
+   GIL is needed to interact with the Python C API).
+
+A reference implementations of the C Exchange API in frameworks:
+
+
+* PyTorch: `C++ <https://github.com/pytorch/pytorch/blob/main/torch/csrc/Module.cpp#L692>`__
```
leofang: Let's link to a particular commit (copy permalink) in case the upstream drifts away.

docs/source/python_spec.rst
```diff
@@ -179,6 +179,42 @@ refer to `github.com/dmlc/dlpack <https://github.com/dmlc/dlpack>`_.
...
+* Paddle: `C++ <https://github.com/PaddlePaddle/Paddle/blob/develop/paddle/fluid/pybind/pybind.cc#L856>`__
```
leofang: Ditto

docs/source/python_spec.rst
```diff
@@ -179,6 +179,42 @@ refer to `github.com/dmlc/dlpack <https://github.com/dmlc/dlpack>`_.
...
+
+.. note:: Implementation of the C Exchange API
+
+   Producer framework should implement the C Exchange API in a static way either
+   through Cython, Python C extensions, or Python binding mechanism. Importantly,
+   because the DLPack C exchange API operates at the C extension level, we need
+   direct interaction between the array framework PyObject* and DLPack,
+   as a result it is harder to implement the C Exchange API through ctypes (because
+   ctypes releases GIL by default and sometimes in non-free-threading environment,
+   GIL is needed to interact with the Python C API).
```
leofang: @dalcinl or @seberg to comment. I am aware that this is to capture the meeting discussion on why ctypes does not work for our purposes, but I am a bit confused by the mention of GIL.

include/dlpack/dlpack.h
```diff
@@ -532,16 +532,17 @@ typedef struct DLPackExchangeAPIHeader {
...
+ * This attribute must be set on the type as a Python PyCapsule
+ * with name "dlpack_exchange_api".
  *
  * A consumer library may use a pattern such as:
  *
  * \code
  *
- * PyObject *api_obj = type(tensor_obj).__c_dlpack_exchange_api__;  // as C-code
- * MyDLPackExchangeAPI *api = PyLong_AsVoidPtr(api_obj);
+ * PyObject *api_obj = type(tensor_obj).__dlpack_c_exchange_api__;  // as C-code
```
leofang: Ditto, now that I take a closer look, this seems odd to me

leofang (changes requested):
Thanks, @tqchen! LGTM overall, left some questions that we should address before merging.

docs/source/python_spec.rst
tqchen: This is not valid C code, but as a pesudo cython like code to query the types, happy to change to C code here

leofang: It'd be nice to give a fully working snippet and not just a pseudo code. As it stands it's quite confusing.

include/dlpack/dlpack.h
tqchen: updated

docs/source/python_spec.rst
tqchen: updated

docs/source/python_spec.rst
tqchen: updated to use full code

tqchen:
@leofang please take another look, all comments are addressed

leofang (approved):
LGTM, thanks @tqchen! 

Maybe let's merge this in a few days? Not sure if @seberg or others wants to take another look.

tqchen:
sounds good, we can wait for a few days

docs/source/python_spec.rst
seberg: Hmmm, this flew past me. Honestly, I am not sure what this note is getting at.  The mention of the GIL seems wrong either way. I think I said it in the meeting too, but don't quite remember.

"Releasing the GIL" actually should be thought of "releases the thread-state". And that affects both free-threaded and non-free threaded python just the same.
The difference is that with the GIL a Python thread must own the thread state (and thus the GIL).

I would be fine with just dropping the note, because I am not sure which functions this affects how right now.
The interesting point to make is that some functions are expected to be called without the thread-state (GIL) held, so they need to grab it (like the note that we have for the deallocator).

seberg:
Yeah, I don't mind (or very slightly prefer) going towards a capsule.

docs/source/python_spec.rst
tqchen: Thanks a lot for the note, updated to replace thread state to GIL

tqchen:
Seems we are close to merge, planning to merge this in incoming week. but happy to also see comments

leofang:
Thanks, Tianqi & all! Let's get this in!

PR #188, opened 2026-06-18
Add a kDLTpu and kDLTPUHost for device and host TPU memory.

pschuh:
This can be used to share buffers between different clients connected to the same TPU runtime like TorchTPU and JAX.

### DIFF
--- a/include/dlpack/dlpack.h
+++ b/include/dlpack/dlpack.h
@@ -120,6 +120,10 @@ typedef enum {
   kDLMAIA = 17,
   /*! \brief AWS Trainium */
   kDLTrn = 18,
+  /*! \brief Google TPU device */
+  kDLTPU = 19,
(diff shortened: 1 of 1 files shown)


tqchen:
Thanks @pschuh , i think TPU makes sense, is there a TPUHost memory in the sense of driver/runtime API?

pschuh:
Hi, yes, this is a mirror of kDLCuda and kDLCudaHost. See https://docs.jax.dev/en/latest/notebooks/host-offloading.html#namedsharding-and-memory-kinds . Like GPU, TPU has faster transfers between the device and pinned_host memory. Our runtime also tracks unpinned_host, but there is nothing special about that memory and it can be kDLCpu.

tqchen (approved):

tqchen:
Thanks @pschuh , this is merged

